Issue #11, opened 2019-12-13
Benchmark framwork 


phamtai97:
Hi all.
Before using the framework I usually care about its performance. Have you tried this framework benchmark? 
You can refer here about benchmark: https://github.com/smallnest/go-web-framework-benchmark

System-Glitch:
Hello, thank you for this. I will try and add the results to the readme and documentation.

phamtai97:
I have some experience with benchmarks service, can I join this project?

System-Glitch:
Sure! Feel free to open a PR.

phamtai97:
Great!!!. I wil try do it.

phamtai97:
Hi @System-Glitch 
What types of routers can Goyave support? Routes such as: "PUT", "DELETE", "OPTION", "GET" have params, ..... 
I see the example route "GET" on the README.

System-Glitch:
All HTTP methods are supported. JSON bodies are supported (don't forget the `Content-Type` header) as well as url-encoded bodies (parsed using Go's standard library).

phamtai97:
Yeah !!! 
I will write a test benchmark soon.

System-Glitch:
@phamtai97 Any news?

phamtai97:
Hi, I am sorry. I've been busy lately. I just review it and do it.

phamtai97:
Hi @System-Glitch. You can help me check pull request [this](https://github.com/System-Glitch/goyave/pull/27).

System-Glitch:
Closed for inactivity.

PR #27, opened 2020-01-10
Benchmark router's goyave.

phamtai97:
## Benchmark router's goyave.
I relied on the idea of ​​the repo [go-http-routing-benchmark](https://github.com/julienschmidt/go-http-routing-benchmark) to perform the benchmark.

## Related issue(s)

* #11


bench_route.go
```diff
@@ -0,0 +1,496 @@
+package goyave
```
System-Glitch: This file should be included inside the test file to avoid being compiled and using memory in applications using the framework.

bench_test.go
```diff
@@ -0,0 +1,114 @@
...
+	runtime.ReadMemStats(m)
+	before := m.HeapAlloc
+
+	load()
+
+	// after
+	runtime.GC()
+	runtime.GC()
+	runtime.GC()
+	runtime.GC()
```
System-Glitch: Why is quadruple manual garbage collection needed?

bench_test.go
```diff
@@ -0,0 +1,114 @@
...
+			router.Route("GET", r.path, goyaveHandler, nil)
+		case "POST":
+			router.Route("POST", r.path, goyaveHandler, nil)
+		case "PATCH":
+			router.Route("PATCH", r.path, goyaveHandler, nil)
+		case "PUT":
+			router.Route("PUT", r.path, goyaveHandler, nil)
+		case "DELETE":
+			router.Route("DELETE", r.path, goyaveHandler, nil)
+		}
```
System-Glitch: This can be shortened.

```go
router.Route(r.method, r.path, goyaveHandler, nil)
```

go.sum
```diff
@@ -5,6 +5,7 @@ cloud.google.com/go v0.37.4/go.mod h1:NHPJ89PdicEuT9hdPXMROBD91xc5uRDxsMtSB16k7h
 github.com/BurntSushi/toml v0.3.1/go.mod h1:xHWCNGjB5oqiDr8zfno3MHue2Ht5sIBksp03qcyfWMU=
 github.com/Shopify/sarama v1.19.0/go.mod h1:FVkBWblsNy7DGZRfXLU0O9RCGt5g3g3yEuWXgklEdEo=
 github.com/Shopify/toxiproxy v2.1.4+incompatible/go.mod h1:OXgGpZ6Cli1/URJOF1DMxUHB2q5Ap20/P/eIdh4G0pI=
+github.com/System-Glitch/goyave v1.0.0 h1:wmDg+BI+VqQ5ifCcjD2j5bQgfLvG/TGPlt6PGBuYv68=
```
System-Glitch: Can you remove this line please?

bench_test.go
phamtai97: yes.

bench_route.go
phamtai97: oh, i will move it to file bench_test.go

bench_test.go
phamtai97: Oh, I just want to make sure GC collects all the garbage

phamtai97:
I do not understand about content "a route with a body (for parsing) and validation."

bench_test.go
System-Glitch: I think it's not very clean. Is there a better way?

System-Glitch:
> I do not understand about content "a route with a body (for parsing) and validation."

You routes are all benched using empty body, meaning that nothing will be parsed. It's only realistic for a GET request, but in a POST request, the data will always be parsed and validated. This parsing and validation process is the most interesting part to benchmark as it is the most resource-intensive one.

bench_test.go
phamtai97: At the moment, I don't have any other way

phamtai97:
Oh thank you. I understand it.

bench_test.go
```diff
@@ -0,0 +1,652 @@
...
+
+	println(""+name+":", after-before, "Bytes")
+}
+
+func goyaveHandler(res *Response, req *Request) {
+	res.String(http.StatusOK, "OK")
+}
+
+func loadGoyaveRouter(routes []route) *Router {
+	config.Load()
```
System-Glitch:
```suggestion
```
If this benchmark is focused on performance once the server is started, config doesn't need to be loaded. Config and language files are only loaded when `goyave.Start()` is called.

bench_test.go
```diff
@@ -0,0 +1,652 @@
...
+			router.requestHandler(w, r, goyaveHandler, nil)
+		}
+	}
+}
+
+func Benchmark_Static(b *testing.B) {
+	println("# Static routes len:", len(staticRoutes))
+	calcMem("Goyave", func() {
+		staticGoyave = loadGoyaveRouter(staticRoutes)
+	})
```
System-Glitch: Loading shouldn't be part measured in the benchmark. Move initialization in an `init()` function.

bench_test.go
```diff
@@ -0,0 +1,652 @@
...
+func Benchmark_GPlus(b *testing.B) {
+	println("# Google+ routes len:", len(gplusAPIRoutes))
+	calcMem("Goyave", func() {
+		staticGoyave = loadGoyaveRouter(gplusAPIRoutes)
+	})
+	benchRoutes(b, staticGoyave, gplusAPIRoutes)
+}
+
+func Benchmark_Validate(b *testing.B) {
+	config.Load()
```
System-Glitch:
```suggestion
```
Loading config is not needed.

System-Glitch:
Closed for inactivity.

frederikhors:
@System-Glitch can you please reopen this?

System-Glitch:
@frederikhors This is no longer relevant. The router has been swapped with a custom implementation long ago and there is already some benchmarks for it.

PR #95, opened 2020-05-03
Make independant CI jobs run in parallel

jRimbault:
## Description

Run each independent continuous integration jobs in parallel, instead of sequentially, one after the other.
I'm not sure about the name for the `ineffassign` check.
The workflow file should maybe be renamed : `test.yml` to something else, `ci.yml` ?

### Possible drawbacks

A bit of duplication in the workflow file ? I'm sure there's a way around that.
Maybe you'll have to check the integration with coveralls.

And since I changed the workflow file, automated requirements for PRs will be a bit thrown off. Possible workarounds : 
- re-add the old tasks, merge, then change PRs' requirements, then remove the old tasks
- or temporarily disable PRs' requirements

## Related issue(s)

None.

_(Really it's just boredom. I promise, next time I'll look at the open issues before)_

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,4 +1,4 @@
-name: Test
+name: CI
 on:
   push:
     branches:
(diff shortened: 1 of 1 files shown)


System-Glitch:
Thank you very much for the contribution! While we're at it, do you know any way to run the same jobs for both Go 1.13 and Go 1.14? (Only measure coverage for 1.14)

PR #118, opened 2020-10-26
Adding extra data to request

### LINKED ISSUE #117: Request extra-data or with context value
## Proposal

It is currently not possible to add extra data to requests other than using `request.Data` or `request.User`. `http.Request` has `WithContext()`, which allows to add a context (`WithValue()`) to it.

Extra data in requests would allow middleware to process data that is not part of the request's body.

### Implementation

The simplest solution would be to add a `map[string]interface{}` named `Extra` to the request structure. Initialize it empty to avoid memory usage.

### Possible drawbacks

`None`.

gmgalvan:
## Description

Make a clear and detailed description of your changes, with the added benefits.

Adding an extra data to Request that would allow middleware to process data that is not part of the request's body
Example a new middleware that could wrap extra information to teh Request and later be processed.

### Possible drawbacks

`None`

## Related issue(s)

List all related issues here:

* closes #117

### DIFF
--- a/docs_src/src/guide/basics/requests.md
+++ b/docs_src/src/guide/basics/requests.md
@@ -487,6 +487,10 @@ fmt.Println(request.Lang) // "en-US"
 fmt.Println(lang.Get(request.Lang, "validation.rules.required")) // "The :field is required."
 ```
 
+#### Request.Extra
+
+`Extra` is a `map[string]interface{}` extra data to request that would allow middleware to process data that is not part of the request's body.
+
 #### Request.User
 
 <p><Badge text="Since v2.5.0"/></p>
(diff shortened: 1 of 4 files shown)


gmgalvan:
Just I have a doubt about where to initialize it empty in order to avoid memory usage and if is needed an example in a build in middleware or new one.

System-Glitch (changes requested):
Currently the map will be nil everywhere, you have to initialize it empty in `router.go#requestHandler()` and `testsuite.go#CreateTestRequest()`.

A quick example in the documentation would be nice too.

gmgalvan:
I added one example about creating a  new middleware and adding the extra information to the request, and I think I initialize empty in router.go#requestHandler() and testsuite.go#CreateTestRequest() the Extra property. mm do we need to update the branch? or something that am I missing?

System-Glitch (approved):

System-Glitch:
Thank you very much!

Issue #105, opened 2020-08-13
Rate limiting


System-Glitch:
## Proposal

To make rate limiting easier, a middleware could be developed. This middleware would use a new config entry making it possible to adjust the rate easily. The real goal is to make a basic rate limiter middleware, which would cover most cases and be easy to use, mostly to protect from attacks.

The middleware will have to correctly set the following headers, following this [IETF spec](https://tools.ietf.org/id/draft-polli-ratelimit-headers-03.html): 
- `RateLimit-Limit`: the rate limit ceiling
- `RateLimit-Remaining`: the number of requests left
- `RateLimit-Reset`: the number of seconds until the quota resets

*Note: the IETF spec doesn't use the `X-` header prefix, unlike popular APIs such as Github's. It would be worth making some research to know if it would be better to add the `X-` prefix or keep it as defined in the IETF spec.*

In case of cancelled request because of limiting, the middleware would be stopping (don't call `next()`) and would respond with the HTTP status `429 Too Many Requests`, in accordance with [`RFC 6585`](https://tools.ietf.org/html/rfc6585#section-4). The default status handler can be kept and developers will be able to define the behavior of their choice when a request is rate limited.

Limiting will be done using an in-memory storage directly implemented in the middleware. A simple map and a struct containing the user's information (IP, time frame, remaining requests) are probably more than enough. The storage **needs** to be **thread-safe**, as requests can be served in parallel.

The official go package [`golang.org/x/time/rate`](https://pkg.go.dev/golang.org/x/time/rate?tab=doc) could be used, but it would need to be extended a little bit to support multiple clients. We want to limit IPs that keep requesting the server, but not the other legitimate users.

I would like to start with a simple rate limiter, as mentioned above, but a more advanced rate limiter could be developed next to it later on. There is no design or clear requirements specification for it yet. For example, this advanced rate limiter could:
- work with authenticated users and add variable limits depending on a user plan
- be configured to count rates in relation to a route
- use an external storage driver instead of the in-memory one explained earlier (support the database for example)
- check the `X-Forwarded-For` and `X-Real-IP` headers if the application is running behind a reverse proxy
- and more?

Either way, this middleware **has to** be compliant with previously mentioned [IETF spec](https://tools.ietf.org/id/draft-polli-ratelimit-headers-03.html).

### Possible drawbacks

Rate limiting can have a slight impact of performance, and especially on memory usage if using in-memory storage. However, it is a mandatory feature for public APIs and well worth the upfront cost.

[...]

PR #120, opened 2020-10-31
add rate limit middleware

agbaraka:
## Description
This pull requests adds a middleware for rate limiting functionality as proposed in [Rate limiting #105](https://github.com/System-Glitch/goyave/issues/105).

The current implementation can:
- Work with authenticated users and add variable limits depending on a user plan
- Be configured to count rates in relation to a route. This can be done by creating multiple middlewares with different configurations.

### Example

```go

// configure the middleware
ratelimiterMiddleware := New(func(request *goyave.Request) LimiterConfig {
    return LimiterConfig {
	RequestQuota:  2,
	QuotaDuration: 10 * time.Minute,
    }
})

// apply to a route
router.Get("/", hello.SayHi).Middleware(ratelimiterMiddleware)
```

### Possible drawbacks

Please note that this is initial implementation, and I am still working to make it compliant to [IETF spec](https://tools.ietf.org/id/draft-polli-ratelimit-headers-03.html). The headers are not yet implemented.

## Related issue(s)

List all related issues here:

* [Rate limiting #105](https://github.com/System-Glitch/goyave/issues/105).

## Final Thoughts

As it is my first time contributing a feature, I would love to get your valuable feedback and opinion.
Also please feel free to provide better naming suggestions :-)


System-Glitch:
As it is still a work in progress, I am converting this PR to draft.

middleware/ratelimiter/ratelimiter.go
```diff
@@ -0,0 +1,88 @@
...
+			key := config.ClientID
+
+			l := lstore.get(key)
+
+			if l == nil {
+				l = newLimiter(config.RequestQuota, config.QuotaDuration)
+				lstore.set(key, l)
+			}
+
+			if time.Since(l.updatedAt) < l.quotaDuration {
```
System-Glitch, line 59: There is a possible race condition here. The `l` variable is not protected by mutex. The call through `lstore` is, though, but you should move that logic to the `limiter` struct.

As it is a pointer, make your changes directly to the `l` variable and don't use `lstore` as a proxy.

middleware/ratelimiter/store.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+func (ls *limiterStore) set(key string, limiter *limiter) {
+	ls.mx.Lock()
+	defer ls.mx.Unlock()
+	ls.store[key] = limiter
+}
+
+func (ls *limiterStore) get(key string) *limiter {
+	ls.mx.Lock()
```
System-Glitch, line 48: Use `RLock()` to benefit from `RWMutex`.

System-Glitch (changes requested):
Good first iteration! Naming is fine and the `middleware/ratelimiter` location is on point.

The configuration function is handy to set the a custom client ID (such as user ID).

I noticed a few problems with the locks and the limiter store that needs to be fixed.

I will make another review when the IETF spec is fully implemented.

middleware/ratelimiter/ratelimiter.go
agbaraka: I am using mutex locks on `lstore` functions such as `set, increments and reset` and none in the `limiter` functions. Making changes directly to `l` variable (without mutex locks), wouldn't it cause race conditions?

Sorry, my knowledge in this area is limited.

System-Glitch: Imagine two requests for the same user occur on two goroutines.

Goroutine 1 locks in `lstore.reset()`, and changes the updateAt value. At the same time, goroutine 2 checks if `time.Since(l.updatedAt) < l.quotaDuration`, but without locking anything. There will be a race condition between goroutine 1 (write) and goroutine 2 (read).

agbaraka: Oh ok. I understand now. I will look into this

System-Glitch:
Note: the IETF draft has been updated. I haven't checked the changes yet.

https://tools.ietf.org/id/draft-polli-ratelimit-headers-04.html

middleware/ratelimiter/ratelimiter.go
```diff
@@ -0,0 +1,101 @@
...
+			l.increment()
+
+			setResponseHeaders(response, l)
+			next(response, request)
+		}
+	}
+}
+
+func setResponseHeaders(response *goyave.Response, l *limiter) {
+	response.Header().Add(
```
System-Glitch, line 74: `Add()` appends to any existing values associated with key. I think it would be preferable to use `Set()` here.

middleware/ratelimiter/ratelimiter.go
```diff
@@ -0,0 +1,101 @@
...
+		"RateLimit-Reset",
+		fmt.Sprintf("%v", l.getSecondsToQuotaReset()),
+	)
+}
+
+func defaultClientID(request *goyave.Request) string {
+	remoteAddr := request.RemoteAddress()
+
+	// strip off port number
+	addrSlice := strings.Split(remoteAddr, ":")
```
System-Glitch, line 94: Use slices and `strings.Index()` instead of `Split()` to avoid allocation.

middleware/ratelimiter/ratelimiter.go
```diff
@@ -0,0 +1,101 @@
...
+		return func(response *goyave.Response, request *goyave.Request) {
+
+			config := configFn(request)
+
+			if config.RequestQuota == 0 || config.QuotaDuration == 0 {
+				next(response, request)
+				return
+			}
+
+			if config.ClientID == "" {
```
System-Glitch, line 44: Zero-value for `interface{}` is `nil`, not `""`.
```suggestion
			if config.ClientID == nil {
```

middleware/ratelimiter/store.go
```diff
@@ -0,0 +1,86 @@
...
+	"sync"
+	"time"
+)
+
+type limiter struct {
+	requestQuota  int
+	counter       int
+	quotaDuration time.Duration
+	resetsAt      time.Time
+	mx            *sync.RWMutex
```
System-Glitch, line 14:
```suggestion
	mx            sync.RWMutex
```
Don't use a pointer for mutex. After this change, you can also remove the initialization from `newLimiter()`

middleware/ratelimiter/store.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+func (l *limiter) getSecondsToQuotaReset() float64 {
+	l.mx.RLock()
+	defer l.mx.RUnlock()
+	return -1 * math.Round(time.Since(l.resetsAt).Seconds())
+}
+
+type limiterStore struct {
+	mx    *sync.RWMutex
```
System-Glitch, line 65:
```suggestion
	mx    sync.RWMutex
```
Same thing, don't use a pointer for mutex.

middleware/ratelimiter/ratelimiter.go
```diff
@@ -0,0 +1,101 @@
...
+
+			if l.hasExpired() {
+				l.reset()
+			} else if l.hasExceededRequestQuota() {
+				setResponseHeaders(response, l)
+				response.Status(http.StatusTooManyRequests)
+				return
+			}
+
+			l.increment()
```
System-Glitch, line 65: It would be better to lock (using limiter's mutex) during this whole operation instead of at a unit level in each `limiter` function. This would avoid unexpected effects, for example when one routine resets the `limiter` just before another goroutine does `setResponseHeaders`, which would result in incoherent values returned to the client. Also that would reduce the number of thread synchronization needed to run this middleware.

System-Glitch (changes requested):
There are still a few things to change before it's perfect, but we're getting close! Concurrency in particular can be improved.

middleware/ratelimiter/store.go
```diff
@@ -0,0 +1,109 @@
...
+
+func (l *limiter) hasExceededRequestQuota() bool {
+	return l.counter >= l.requestQuota
+}
+
+func (l *limiter) getRemainingRequestQuota() int {
+	return l.requestQuota - l.counter
+}
+
+func (l *limiter) getSecondsToQuotaReset() float64 {
```
System-Glitch, line 84: Shouldn't it return an int?

middleware/ratelimiter/store.go
```diff
@@ -0,0 +1,109 @@
...
+		quotaDuration: quotaDuration,
+		counter:       0,
+		resetsAt:      time.Now().Add(quotaDuration),
+	}
+}
+
+func (l *limiter) validateAndUpdate(response *goyave.Response) bool {
+
+	l.mx.RLock()
+	defer l.mx.RUnlock()
```
System-Glitch, line 32:
```suggestion
	l.mx.Lock()
	defer l.mx.Unlock()
```

middleware/ratelimiter/store.go
```diff
@@ -0,0 +1,109 @@
...
+
+	"github.com/System-Glitch/goyave/v3"
+)
+
+type limiter struct {
+	requestQuota  int
+	counter       int
+	quotaDuration time.Duration
+	resetsAt      time.Time
+	mx            sync.RWMutex
```
System-Glitch, line 17:
```suggestion
	mx            sync.Mutex
```
Only `newLimiter` uses the mutex now, and it both reads and write. Therefore, we can use a regular mutex.

middleware/ratelimiter/ratelimiter_test.go
```diff
@@ -0,0 +1,111 @@
...
+
+	suite.Equal(
+		fmt.Sprintf("%v, %v;w=%v", requestQuota, requestQuota, quotaDuration.Seconds()),
+		result.Header.Get("RateLimit-Limit"),
+	)
+
+	suite.Equal(
+		fmt.Sprintf("%v", requestQuota-numberOfRequests),
+		result.Header.Get("RateLimit-Remaining"),
+	)
```
System-Glitch, line 53: `RateLimite-Reset` header's value is not tested.

System-Glitch (changes requested):
Some tiny changes left to do. Otherwise, it looks all good to me! 👍
I will pull this on my machine and test it before merging.

middleware/ratelimiter/store.go
agbaraka: Oh ok... noted. Thank you

middleware/ratelimiter/ratelimiter_test.go
agbaraka: Oh yes! I have been trying to figure out how to test it.

I would like to get your input on this.

middleware/ratelimiter/store.go
agbaraka: I used float64 only because `math.Round` returns float64, and I had no straight forward reason to convert it to int.

Kindly advise

System-Glitch: Right, `Sprintf` with `%v` will not show the decimal places anyway if there are none.

middleware/ratelimiter/ratelimiter_test.go
System-Glitch: - Get the remote address from the request created with `suite.CreateTestRequest`.
- Use it to get the instance of `limiter` in your store.
- Inject your own newly generated time so you can compare easily without the risk of time difference due to execution time.

middleware/ratelimiter/ratelimiter.go
```diff
@@ -19,6 +19,8 @@ type LimiterConfig struct {
 	// Unique identifier for requestors. Can be userID or IP
 	// Defaults to Remote Address if it is empty
 	ClientID interface{}
+
+	secondsToQuotaReset float64
```
System-Glitch, line 23: This is not what I meant. What you've done is a workaround only used for testing. It defeats the whole purpose of testing.

Here is an example of what I was thinking about:
```go
request := suite.CreateTestRequest(nil)
clientID := request.RemoteAddress()
resetTime := time.Now().Add(5 * time.Second)

store := newLimiterStore()
limiter := store.get(clientID)
limiter.resetsAt = resetTime

middleware := newWithStore(store)
result := suite.Middleware(...)
suite.Equal(resetTime, result.Header.Get("RateLimit-Reset")
```

This would need some very light refactoring so we're able to pass a custom limiter store:
```go
func New(configFn LimiterConfigFunc) goyave.Middleware {
	return newWithStore(configFn, newLimiterStore())
}

func newWithStore(configFn LimiterConfigFunc, lstore limiterStore) goyave.Middleware {
	return func(next goyave.Handler) goyave.Handler {
		return func(response *goyave.Response, request *goyave.Request) {
			//...
		}
	}
}
```

System-Glitch (changes requested):

agbaraka:
Hi @System-Glitch. 

You were right with regards to the above comment. It was something I even had to convince myself on it's validity.

As suggested, I have made some code refactoring. Please check and advise

Thank you!

System-Glitch:
> 
> 
> Hi @System-Glitch.
> 
> You were right with regards to the above comment. It was something I even had to convince myself on its validity.
> 
> As suggested, I have made some code refactoring. Please check and advise
> 
> Thank you!

Thank you for the change! Everything looks fine to me now. I will just test it on my end before approval.

System-Glitch:
I added a few unit tests and added your name to the contributors list. Thank you very much for your contribution! 🎉👍

System-Glitch (approved):

System-Glitch:
@agbaraka While writing the docs, I thought about a potential Out of Memory problem. If the server runs for a long time and a lot of different clients are requesting the server, the limiterStore map will grow huge and never shrink.

I will try to find a way to periodically clear this map from entries for which the quota has reset.

agbaraka:
@System-Glitch That's true. I did not think of that scenario. Please let me know once you have merged a solution to this issue

System-Glitch:
@agbaraka 
Implementation idea to address the memory issue:

- When creating a new limiter, also use `time.AfterFunc()` to call a function removing it from the limiter store.
- The runtime implementation being very efficient, this shouldn't cause a big impact on performance. Using this approach wouldn't require to lock and iterate on the whole store on a regular basis, which makes it very much easier to implement and also way more efficient.
- Remove `limiter.reset()` and `limiter.resetsAt`, as expired limiters will never be part of the store anymore.
- This would require a bit of refactoring so removal can be done in a concurrently safe way.

Unrelated change but I wanted to notice you about it: I will also increment the request counter when the quota is exceeded. This allows better potential monitoring of excessive request rate.

agbaraka:
@System-Glitch 

It looks good to me.

The only concern I have is on the store map memory management.From your comment *Warning though, go maps aren't shrunk after key deletion*, since the map does not shrink, is deleting items have the same impact in the memory as keeping them around? Can this be mitigating by using embedded stores like Badger?

System-Glitch:
@agbaraka Deleting items still has a good impact on memory, since the already allocated space in the map can be re-used for other clients. It would still grow very large if a lot of client all request the API in the same time-span (which is the quota duration). To limit this problem, it is best to use short quota durations (such as "per-minute" quotas).

I would like to avoid importing another library for this single use. At any rate, this implementation is sufficient for the goal we are reaching for in my opinion: providing a simple rate limiting solution. Huge applications with huge traffic will probably use their own implementation anyway so they can also limit based on behavior, bot detection algorithms, etc.

agbaraka:
@System-Glitch Thanks for the explanation.

Everything is now ok on my end.

Issue #138, opened 2021-03-29
Support JWT custom claims


Morishiri:
## Proposal

I propose to support custom claims in the generated JWT token.

The user fields to be included as claims could be specified by adding `auth:"jwtInclude"` (to follow current `auth` prefix).

This would allow people putting custom information in the token, e.g. when thinking about username, some additional name fields or permissions (I saw you're planning support in the future).

### Possible drawbacks

By default - none, if someone uses too many fields then the token will be heavy, but that's up to user of the framework.

## Additional information

I think I can handle this contribution if you agree.

System-Glitch:
Hello and thank you for your proposal! I was planning on adding a `auth.GenerateTokenWithClaims()` function in the near future, which would let you add custom claims.

Can you elaborate on how you planned on implementing this feature? I think this would be a great addition.

Do you need your JWT in other applications? You probably don't need to add anything else than the ID of the user since the authenticator will fetch the user for you in the database, so I'm not sure a struct tag `auth:"jwtInclude"` would be really useful (and it would require some reflection).

Morishiri:
> I was planning on adding a auth.GenerateTokenWithClaims() function in the near future, which would let you add custom claims.
> 
> Can you elaborate on how you planned on implementing this feature? I think this would be a great addition.

Adding `auth.GenerateTokenWithClaims()` would help, my plan was simply not to introduce additional functions, but it actually may be better to do so to not use reflection too much. The idea was to add `auth:"jwtInclude"` annotation to the fields from the User struct which should be added to the token, then get this values from the user object and simply put them as claims, but as you already said, this would require reflection to get the fields annotations and may slow things down.

> Do you need your JWT in other applications? You probably don't need to add anything else than the ID of the user since the authenticator will fetch the user for you in the database, so I'm not sure a struct tag auth:"jwtInclude" would be really useful (and it would require some reflection).

It can be used for e.g. to put user first and last name to the token, so when the user is logged in I can simply look into the token and display his name in the header menu (e.g. avatar button on github, there is `Signed in as <nickname>` text) without the need to execute a call to `/users/me` (or some similar route). But it's not critical. The authenticator fetching the user object is really great feature and implementing such `/me` endpoint is not a big hassle with that.

What I proposed was mainly front-end related to just read the token and not call APIs, may be improvement for some people, but I agree that in this case separate function may be better approach.

System-Glitch:
[...]

PR #139, opened 2021-04-01
Support custom JWT claims

Morishiri:
## Description

Support for generating JWT token with custom claims. Discussed in #138. Will open docs PR for documentation update.

### Possible drawbacks

`None`

## Related issue(s)

* #138 

## Additional information

Also includes test changes, splitting them and making cleaner.

### DIFF
--- a/auth/jwt_controller.go
+++ b/auth/jwt_controller.go
@@ -22,15 +22,7 @@ import (
 // The token is set to expire in the amount of seconds defined by
 // the "auth.jwt.expiry" config entry.
 func GenerateToken(username interface{}) (string, error) {
-	expiry := time.Duration(config.GetInt("auth.jwt.expiry")) * time.Second
-	now := time.Now()
-	token := jwt.NewWithClaims(jwt.SigningMethodHS256, jwt.MapClaims{
-		"userid": username,
-		"nbf":    now.Unix(),             // Not Before
-		"exp":    now.Add(expiry).Unix(), // Expiry
-	})
-
-	return token.SignedString([]byte(config.GetString("auth.jwt.secret")))
+	return GenerateTokenWithClaims(jwt.MapClaims{"userid": username})
 }
 
 // GenerateTokenWithClaims generates a new JWT with custom claims.
@@ -38,13 +30,16 @@ func GenerateToken(username interface{}) (string, error) {
 // the "auth.jwt.secret" config entry.
 // The token is set to expire in the amount of seconds defined by
 // the "auth.jwt.expiry" config entry.
-func GenerateTokenWithClaims(username interface{}, claims jwt.MapClaims) (string, error) {
+func GenerateTokenWithClaims(claims jwt.MapClaims) (string, error) {
 	expiry := time.Duration(config.GetInt("auth.jwt.expiry")) * time.Second
 	now := time.Now()
-	customClaims := claims
-	customClaims["userid"] = username
-	customClaims["nbf"] = now.Unix()             // Not Before
-	customClaims["exp"] = now.Add(expiry).Unix() // Expiry
+	customClaims := jwt.MapClaims{
+		"nbf": now.Unix(),             // Not Before
+		"exp": now.Add(expiry).Unix(), // Expiry
+	}
+	for k, c := range claims {
+		customClaims[k] = c
+	}
 	token := jwt.NewWithClaims(jwt.SigningMethodHS256, customClaims)
 
 	return token.SignedString([]byte(config.GetString("auth.jwt.secret")))
@@ -52,35 +47,33 @@ func GenerateTokenWithClaims(username interface{}, claims jwt.MapClaims) (string
 
 // TokenFunc is the function used by JWTController to generate tokens
 // during login process.
-type TokenFunc func(username interface{}, user interface{}) (string, error)
+type TokenFunc func(request *goyave.Request, user interface{}) (string, error)
 
 // JWTController a controller for JWT-based authentication, using HMAC SHA256.
 // Its model fields are used for username and password retrieval.
 type JWTController struct {
-	model interface{}
-
+	model     interface{}
+	TokenFunc TokenFunc
 	// UsernameField the name of the request's body field
 	// used as username in the authentication process
 	UsernameField string
-
 	// PasswordField the name of the request's body field
 	// used as password in the authentication process
 	PasswordField string
-
-	TokenFunc TokenFunc
 }
 
 // NewJWTController create a new JWTController that will
 // be using the given model for login and token generation.
 func NewJWTController(model interface{}) *JWTController {
-	return &JWTController{
+	controller := &JWTController{
 		model:         model,
 		UsernameField: "username",
 		PasswordField: "password",
-		TokenFunc: func(username interface{}, user interface{}) (string, error) {
-			return GenerateToken(username)
-		},
 	}
+	controller.TokenFunc = func(r *goyave.Request, user interface{}) (string, error) {
+		return GenerateToken(r.String(controller.UsernameField))
+	}
+	return controller
 }
 
 // Login POST handler for token-based authentication.
@@ -106,7 +99,7 @@ func (c *JWTController) Login(response *goyave.Response, request *goyave.Request
 
 	pass := reflect.Indirect(reflect.ValueOf(user)).FieldByName(columns[1].Field.Name)
 	if !notFound && bcrypt.CompareHashAndPassword([]byte(pass.String()), []byte(request.String(c.PasswordField))) == nil {
-		token, err := c.TokenFunc(username, user)
+		token, err := c.TokenFunc(request, user)
 		if err != nil {
 			panic(err)
 		}
--- a/auth/jwt_controller.go
+++ b/auth/jwt_controller.go
@@ -21,17 +21,38 @@ import (
 // the "auth.jwt.secret" config entry.
 // The token is set to expire in the amount of seconds defined by
 // the "auth.jwt.expiry" config entry.
-func GenerateToken(id interface{}) (string, error) {
+func GenerateToken(username interface{}) (string, error) {
 	expiry := time.Duration(config.GetInt("auth.jwt.expiry")) * time.Second
 	now := time.Now()
 	token := jwt.NewWithClaims(jwt.SigningMethodHS256, jwt.MapClaims{
-		"userid": id,
+		"userid": username,
 		"nbf":    now.Unix(),             // Not Before
 		"exp":    now.Add(expiry).Unix(), // Expiry
 	})
 
 	return token.SignedString([]byte(config.GetString("auth.jwt.secret")))
 }
+
+// GenerateTokenWithClaims generates a new JWT with custom claims.
+// The token is created using the HMAC SHA256 method and signed using
+// the "auth.jwt.secret" config entry.
+// The token is set to expire in the amount of seconds defined by
+// the "auth.jwt.expiry" config entry.
+func GenerateTokenWithClaims(username interface{}, claims jwt.MapClaims) (string, error) {
+	expiry := time.Duration(config.GetInt("auth.jwt.expiry")) * time.Second
+	now := time.Now()
+	customClaims := claims
+	customClaims["userid"] = username
+	customClaims["nbf"] = now.Unix()             // Not Before
+	customClaims["exp"] = now.Add(expiry).Unix() // Expiry
+	token := jwt.NewWithClaims(jwt.SigningMethodHS256, customClaims)
+
+	return token.SignedString([]byte(config.GetString("auth.jwt.secret")))
+}
+
+// TokenFunc is the function used by JWTController to generate tokens
+// during login process.
+type TokenFunc func(username interface{}, user interface{}) (string, error)
 
 // JWTController a controller for JWT-based authentication, using HMAC SHA256.
 // Its model fields are used for username and password retrieval.
@@ -45,6 +66,8 @@ type JWTController struct {
 	// PasswordField the name of the request's body field
 	// used as password in the authentication process
 	PasswordField string
+
+	TokenFunc TokenFunc
 }
 
 // NewJWTController create a new JWTController that will
@@ -54,6 +77,9 @@ func NewJWTController(model interface{}) *JWTController {
 		model:         model,
 		UsernameField: "username",
 		PasswordField: "password",
+		TokenFunc: func(username interface{}, user interface{}) (string, error) {
+			return GenerateToken(username)
+		},
 	}
 }
 
@@ -80,7 +106,7 @@ func (c *JWTController) Login(response *goyave.Response, request *goyave.Request
 
 	pass := reflect.Indirect(reflect.ValueOf(user)).FieldByName(columns[1].Field.Name)
 	if !notFound && bcrypt.CompareHashAndPassword([]byte(pass.String()), []byte(request.String(c.PasswordField))) == nil {
-		token, err := GenerateToken(username)
+		token, err := c.TokenFunc(username, user)
 		if err != nil {
 			panic(err)
 		}
--- a/auth/jwt_controller_test.go
+++ b/auth/jwt_controller_test.go
@@ -3,6 +3,7 @@ package auth
 import (
 	"bytes"
 	"encoding/json"
+	"fmt"
 	"net/http"
 	"net/http/httptest"
 	"strings"
@@ -19,6 +20,7 @@ import (
 const testUserPassword = "secret"
 
 type JWTControllerTestSuite struct {
+	user *TestUser
 	goyave.TestSuite
 }
 
@@ -35,12 +37,12 @@ func (suite *JWTControllerTestSuite) SetupTest() {
 	if err != nil {
 		panic(err)
 	}
-	user := &TestUser{
+	suite.user = &TestUser{
 		Name:     "Admin",
 		Email:    "johndoe@example.org",
 		Password: string(password),
 	}
-	database.GetConnection().Create(user)
+	database.GetConnection().Create(suite.user)
 }
 
 func (suite *JWTControllerTestSuite) TestLogin() {
@@ -96,9 +98,10 @@ func (suite *JWTControllerTestSuite) TestLogin() {
 func (suite *JWTControllerTestSuite) TestLoginWithCustomTokenFunc() {
 	controller := NewJWTController(&TestUser{})
 	suite.NotNil(controller)
-	controller.TokenFunc = func(username interface{}, user interface{}) (string, error) {
-		return GenerateTokenWithClaims(username, jwt.MapClaims{
-			"sub": (user.(*TestUser)).ID,
+	controller.TokenFunc = func(r *goyave.Request, user interface{}) (string, error) {
+		return GenerateTokenWithClaims(jwt.MapClaims{
+			"userid": (user.(*TestUser)).Email,
+			"sub":    (user.(*TestUser)).ID,
 		})
 	}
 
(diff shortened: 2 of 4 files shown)


auth/jwt_controller.go
```diff
@@ -21,18 +21,39 @@ import (
 // the "auth.jwt.secret" config entry.
 // The token is set to expire in the amount of seconds defined by
 // the "auth.jwt.expiry" config entry.
-func GenerateToken(id interface{}) (string, error) {
+func GenerateToken(username interface{}) (string, error) {
```
Morishiri, line 24: I changed the parameter name here, because what we really expect is `username` field of user model. Not `id`. It was confusing IMHO.

auth/jwt_controller.go
```diff
@@ -21,18 +21,39 @@ import (
...
+	customClaims["nbf"] = now.Unix()             // Not Before
+	customClaims["exp"] = now.Add(expiry).Unix() // Expiry
+	token := jwt.NewWithClaims(jwt.SigningMethodHS256, customClaims)
+
+	return token.SignedString([]byte(config.GetString("auth.jwt.secret")))
+}
+
+// TokenFunc is the function used by JWTController to generate tokens
+// during login process.
+type TokenFunc func(username interface{}, user interface{}) (string, error)
```
System-Glitch, line 55: First parameter should be `*goyave.Request`.
```suggestion
type TokenFunc func(request *goyave.Request, user interface{}) (string, error)
```

auth/jwt_controller.go
```diff
@@ -80,7 +106,7 @@ func (c *JWTController) Login(response *goyave.Response, request *goyave.Request
 
 	pass := reflect.Indirect(reflect.ValueOf(user)).FieldByName(columns[1].Field.Name)
 	if !notFound && bcrypt.CompareHashAndPassword([]byte(pass.String()), []byte(request.String(c.PasswordField))) == nil {
-		token, err := GenerateToken(username)
+		token, err := c.TokenFunc(username, user)
```
System-Glitch, line 109:
```suggestion
		token, err := c.TokenFunc(request, user)
```

auth/jwt_test.go
```diff
@@ -13,6 +13,8 @@ import (
 
 type JWTAuthenticatorTestSuite struct {
 	goyave.TestSuite
+	tokenAuthenticator *JWTAuthenticator
+	user               *TestUser
```
System-Glitch, line 17: We can afford duplication in tests, it's not a problem. I would prefer to re-define those in all tests in case we need different parameters. (Like in TestOptional) This also ensures with don't have tests overflowing on others.

auth/jwt_test.go
```diff
@@ -47,48 +50,134 @@ func (suite *JWTAuthenticatorTestSuite) createWrongToken(method jwt.SigningMetho
 	return token.SignedString([]byte(config.GetString("auth.jwt.secret")))
 }
 
-func (suite *JWTAuthenticatorTestSuite) TestAuthenticate() {
-	user := &TestUser{}
-	tokenAuthenticator := &JWTAuthenticator{}
+func (suite *JWTAuthenticatorTestSuite) TestGenerateToken() {
```
System-Glitch, line 53:
```suggestion
func (suite *JWTAuthenticatorTestSuite) TestAuthenticate() {
```

I think it is better to name it that way because what we actually want to know is if that Authenticator is working properly. The `GenerateToken` function is tested more in-depth in other test functions in the suite.

auth/jwt_test.go
```diff
@@ -47,48 +50,134 @@ func (suite *JWTAuthenticatorTestSuite) createWrongToken(method jwt.SigningMetho
...
+	suite.Nil(err)
+}
+
+func (suite *JWTAuthenticatorTestSuite) TestGenerateTokenHasClaims() {
+	token, err := GenerateToken("johndoe@example.org")
+	suite.Nil(err)
+	claims := jwt.MapClaims{}
+	_, err = jwt.ParseWithClaims(token, claims, func(token *jwt.Token) (interface{}, error) {
+		return []byte(config.GetString("auth.jwt.secret")), nil
+	})
```
System-Glitch, line 76: Should check signing method as well
```suggestion
	_, err = jwt.ParseWithClaims(token, claims, func(token *jwt.Token) (interface{}, error) {
		if _, ok := token.Method.(*jwt.SigningMethodHMAC); !ok {
			return nil, fmt.Errorf("Unexpected signing method: %v", token.Header["alg"])
		}
		return []byte(config.GetString("auth.jwt.secret")), nil
	})
```

auth/jwt_test.go
```diff
@@ -47,48 +50,134 @@ func (suite *JWTAuthenticatorTestSuite) createWrongToken(method jwt.SigningMetho
...
+
+func (suite *JWTAuthenticatorTestSuite) TestGenerateTokenValidates() {
+	token, err := GenerateToken("johndoe@example.org")
+	suite.Nil(err)
+	claims := jwt.MapClaims{}
+	_, err = jwt.ParseWithClaims(token, claims, func(token *jwt.Token) (interface{}, error) {
+		return []byte(config.GetString("auth.jwt.secret")), nil
+	})
+	suite.Nil(err)
+}
```
System-Glitch, line 68: This test is redundant with the next one ( `TestGenerateTokenHasClaims` )

auth/jwt_test.go
```diff
@@ -47,48 +50,134 @@ func (suite *JWTAuthenticatorTestSuite) createWrongToken(method jwt.SigningMetho
...
+
+	useridPresent := false
+	for key, val := range claims {
+		if key == "userid" {
+			suite.Equal("johndoe@example.org", val)
+			useridPresent = true
+		}
+	}
+
+	suite.True(useridPresent)
```
System-Glitch, line 87:
```suggestion
	userID, ok := claims["userid"]
	suite.True(ok)
	suite.Equal("johndoe@example.org", userID)
```

auth/jwt_test.go
```diff
@@ -47,48 +50,134 @@ func (suite *JWTAuthenticatorTestSuite) createWrongToken(method jwt.SigningMetho
...
+			suite.Equal("johndoe@example.org", val)
+			useridPresent = true
+		}
+		if key == "sub" {
+			suite.Equal(suite.user.ID, uint(val.(float64)))
+			subPresent = true
+		}
+	}
+	suite.True(useridPresent)
+	suite.True(subPresent)
```
System-Glitch, line 114:
```suggestion
	userID, okID := claims["userid"]
	suite.True(okID)
	suite.Equal("johndoe@example.org", userID)
	sub, okSub := claims["sub"]
	suite.True(okSub)
	suite.Equal(suite.user.ID, uint(sub.(float64)))
```

auth/jwt_test.go
```diff
@@ -47,48 +50,134 @@ func (suite *JWTAuthenticatorTestSuite) createWrongToken(method jwt.SigningMetho
...
+}
+
+func (suite *JWTAuthenticatorTestSuite) TestGenerateTokenWithClaimsValidates() {
```
System-Glitch, line 117: This test is redundant with `TestGenerateTokenWithClaimsHasClaims()`  (and tests `GenerateToken` instand of `GenerateTokenWithClaims`

auth/jwt_test.go
```diff
@@ -47,48 +50,134 @@ func (suite *JWTAuthenticatorTestSuite) createWrongToken(method jwt.SigningMetho
...
+	token, err := GenerateToken("johndoe@example.org")
+	suite.Nil(err)
+	claims := jwt.MapClaims{}
+	_, err = jwt.ParseWithClaims(token, claims, func(token *jwt.Token) (interface{}, error) {
+		return []byte(config.GetString("auth.jwt.secret")), nil
+	})
+	suite.Nil(err)
+}
+
+func (suite *JWTAuthenticatorTestSuite) TestGenerateTokenWithClaims() {
```
System-Glitch, line 127:
```suggestion
func (suite *JWTAuthenticatorTestSuite) TestAuthenticateWithClaims() {
```

auth/jwt_controller_test.go
```diff
@@ -92,6 +93,61 @@ func (suite *JWTControllerTestSuite) TestLogin() {
 	result.Body.Close()
 }
 
+func (suite *JWTControllerTestSuite) TestLoginWithCustomTokenFunc() {
+	controller := NewJWTController(&TestUser{})
+	suite.NotNil(controller)
+	controller.TokenFunc = func(username interface{}, user interface{}) (string, error) {
+		return GenerateTokenWithClaims(username, jwt.MapClaims{
```
System-Glitch, line 100:
```suggestion
	controller.TokenFunc = func(r *goyave.Request, user interface{}) (string, error) {
		return GenerateTokenWithClaims(r.String(controller.UsernameField), jwt.MapClaims{
```

auth/jwt_controller.go
```diff
@@ -54,6 +77,9 @@ func NewJWTController(model interface{}) *JWTController {
 		model:         model,
 		UsernameField: "username",
 		PasswordField: "password",
+		TokenFunc: func(username interface{}, user interface{}) (string, error) {
```
System-Glitch, line 80: Create controller before returning, then set the `TokenFunc`. That lets you use the controller's struct fields inside the function's scope.

```go
func NewJWTController(model interface{}) *JWTController {
	controller := &JWTController{
		model:         model,
		UsernameField: "username",
		PasswordField: "password"
	}
	controller.TokenFunc = func(r *goyave.Request, user interface{}) (string, error) {
		return GenerateToken(r.String(controller.UsernameField)
	}
	return controller
}

**Review by System-Glitch — CHANGES_REQUESTED** [review]
Thank you very much for the PR. I appreciate the extra work put into fragmenting the tests to avoid one huge test function. I spotted some things that could be improved, have a look! :)

File: auth/jwt_controller.go (hunk 1 above)
Comment (System-Glitch) [reply auth/jwt_controller.go:24]: Good initiative

File: run_test.sh [hunk 14]
```diff
@@ -42,13 +42,13 @@ echo -e "\033[92m\033[1mDatabase ready. Running tests...\033[0m"
 gcc --version >/dev/null 2>&1
 if [ $? -ne 0 ]; then
 	echo -e "\033[33mgcc is missing. Running tests without data race checking.\033[0m"
-	gotest -v -p 1 -coverprofile=c.out -coverpkg=./... ./... ; go tool cover -html=c.out ; go tool cover -func=c.out | grep total ; rm c.out
+	go test -v -p 1 -coverprofile=c.out -coverpkg=./... ./... ; go tool cover -html=c.out ; go tool cover -func=c.out | grep total ; rm c.out
 else
-	gotest -v -p 1 -race -coverprofile=c.out -coverpkg=./... ./... ; go tool cover -html=c.out ; go tool cover -func=c.out | grep total ; rm c.out
+	go test -v -p 1 -race -coverprofile=c.out -coverpkg=./... ./... ; go tool cover -html=c.out ; go tool cover -func=c.out | grep total ; rm c.out
```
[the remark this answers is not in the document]
Morishiri: https://github.com/rakyll/gotest -> the curl gives 404. I assumed it's no longer maintained. But with go install maybe it will work ;p I will remove it from this PR.

Morishiri:
@System-Glitch please take a look again, I hope I didn't miss any of your comments :)

auth/jwt_controller.go
```diff
@@ -21,18 +21,39 @@ import (
...
 
 	return token.SignedString([]byte(config.GetString("auth.jwt.secret")))
 }
 
+// GenerateTokenWithClaims generates a new JWT with custom claims.
+// The token is created using the HMAC SHA256 method and signed using
+// the "auth.jwt.secret" config entry.
+// The token is set to expire in the amount of seconds defined by
+// the "auth.jwt.expiry" config entry.
+func GenerateTokenWithClaims(username interface{}, claims jwt.MapClaims) (string, error) {
```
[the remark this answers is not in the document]
Morishiri: I just realized, developers may want to name it differently, but we are using it to fetch user from db.
```
result := database.GetConnection().Where(column.Name+" = ?", claims["userid"]).First(user)
```
(https://github.com/go-goyave/goyave/blob/34861e6d93255a6ca61b43567ad192c5b2c7859b/auth/jwt.go#L71)

I think we should either keep it or provide some way to specify fetching algorithm :X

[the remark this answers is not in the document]
Morishiri: Maybe we should instead introduce `sub` claim which is the standard one (https://tools.ietf.org/html/rfc7519#section-4.1)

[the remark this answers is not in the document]
System-Glitch: Interesting, I didn't know about this RFC. At first I was thinking about adding a field to `JWTAuthenticator` to allow users to customize the name of the claim used.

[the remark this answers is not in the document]
System-Glitch: Using the "sub" claim looks better than "userid", but I don't know what kind of effect it would have on developers relying on the feature if we changed the claim name used. If they are using the token outside of their Goyave application, it could break things. They are not really supposed to do that with the current implementation, but it's a possibility. What do you think?

auth/jwt_test.go
```diff
@@ -48,19 +41,120 @@ func (suite *JWTAuthenticatorTestSuite) createWrongToken(method jwt.SigningMetho
...
+	database.GetConnection().Create(user)
+	token, err := GenerateTokenWithClaims(jwt.MapClaims{
+		"sub":    user.ID,
+		"userid": user.Email,
+	})
+	suite.Nil(err)
+	claims := jwt.MapClaims{}
+	_, err = jwt.ParseWithClaims(token, claims, func(token *jwt.Token) (interface{}, error) {
+		return []byte(config.GetString("auth.jwt.secret")), nil
+	})
```
System-Glitch, line 87:
```suggestion
	_, err = jwt.ParseWithClaims(token, claims, func(token *jwt.Token) (interface{}, error) {
		if _, ok := token.Method.(*jwt.SigningMethodHMAC); !ok {
			return nil, fmt.Errorf("Unexpected signing method: %v", token.Header["alg"])
		}
		return []byte(config.GetString("auth.jwt.secret")), nil
	})
```
Should check signing method as well

auth/jwt_test.go
```diff
@@ -48,19 +41,120 @@ func (suite *JWTAuthenticatorTestSuite) createWrongToken(method jwt.SigningMetho
...
-	token, err = GenerateToken("wrongemail@example.org")
+func (suite *JWTAuthenticatorTestSuite) TestTokenWithClaimsValidates() {
+	token, err := GenerateToken("johndoe@example.org")
+	suite.Nil(err)
+	claims := jwt.MapClaims{}
+	_, err = jwt.ParseWithClaims(token, claims, func(token *jwt.Token) (interface{}, error) {
+		return []byte(config.GetString("auth.jwt.secret")), nil
+	})
+	suite.Nil(err)
+}
```
System-Glitch, line 106: This test is redundant with `TestTokenHasClaims()`, you can remove it.

auth/jwt_test.go
```diff
@@ -48,19 +41,120 @@ func (suite *JWTAuthenticatorTestSuite) createWrongToken(method jwt.SigningMetho
...
+
+func (suite *JWTAuthenticatorTestSuite) TestAuthenticateWithClaims() {
+	tokenAuthenticator := &JWTAuthenticator{}
+	user := &TestUser{
+		Name:  "Admin",
+		Email: "johndoe@example.org",
+	}
+	database.GetConnection().Create(user)
```
System-Glitch, line 114:
```suggestion
	user := &TestUser{}
```
See my comment above (same reason)

auth/jwt_test.go
```diff
@@ -48,19 +41,120 @@ func (suite *JWTAuthenticatorTestSuite) createWrongToken(method jwt.SigningMetho
...
+	suite.Equal("Admin", user.Name)
+}
+
+func (suite *JWTAuthenticatorTestSuite) TestGenerateTokenInvalidCredentials() {
+	tokenAuthenticator := &JWTAuthenticator{}
+	user := &TestUser{
+		Name:  "Admin",
+		Email: "johndoe@example.org",
+	}
+	database.GetConnection().Create(user)
```
System-Glitch, line 130:
```suggestion
	user := &TestUser{}
```
See my comment above (same reason)

auth/jwt_test.go
```diff
@@ -48,19 +41,120 @@ func (suite *JWTAuthenticatorTestSuite) createWrongToken(method jwt.SigningMetho
...
+	token, err := GenerateToken("wrongemail@example.org")
 	suite.Nil(err)
 	suite.Equal("These credentials don't match our records.", tokenAuthenticator.Authenticate(suite.createRequest(token), user).Error())
+	suite.Nil(err)
```
System-Glitch, line 134: Extra `suite.Nil(err)` here. You can remove it.

auth/jwt_test.go
```diff
@@ -48,19 +41,120 @@ func (suite *JWTAuthenticatorTestSuite) createWrongToken(method jwt.SigningMetho
...
+}
+func (suite *JWTAuthenticatorTestSuite) TestGenerateTokenWithClaimsInvalidCredentials() {
+	tokenAuthenticator := &JWTAuthenticator{}
+	user := &TestUser{
+		Name:  "Admin",
+		Email: "johndoe@example.org",
+	}
+	database.GetConnection().Create(user)
```
System-Glitch, line 142:
```suggestion
	user := &TestUser{}
```
See my comment above (same reason)

auth/jwt_test.go
```diff
@@ -48,19 +41,120 @@ func (suite *JWTAuthenticatorTestSuite) createWrongToken(method jwt.SigningMetho
...
+}
 
-	user = &TestUser{}
+func (suite *JWTAuthenticatorTestSuite) TestAuthenticateInvalidToken() {
+	tokenAuthenticator := &JWTAuthenticator{}
+	user := &TestUser{
+		Name:  "Admin",
+		Email: "johndoe@example.org",
+	}
+	database.GetConnection().Create(user)
```
System-Glitch, line 157:
```suggestion
	user := &TestUser{}
```
See my comment above (same reason)

auth/jwt_test.go
```diff
@@ -70,23 +164,35 @@ func (suite *JWTAuthenticatorTestSuite) TestAuthenticate() {
...
 	})
+}
 
-	user = &TestUser{}
+func (suite *JWTAuthenticatorTestSuite) TestAuthenticateTokenInFuture() {
+	tokenAuthenticator := &JWTAuthenticator{}
+	user := &TestUser{
+		Name:  "Admin",
+		Email: "johndoe@example.org",
+	}
```
System-Glitch, line 180:
```suggestion
	user := &TestUser{}
```

auth/jwt_test.go
```diff
@@ -70,23 +164,35 @@ func (suite *JWTAuthenticatorTestSuite) TestAuthenticate() {
...
+}
 
-	user = &TestUser{}
-	nbf = time.Now()
+func (suite *JWTAuthenticatorTestSuite) TestAuthenticateTokenExpired() {
+	tokenAuthenticator := &JWTAuthenticator{}
+	user := &TestUser{
+		Name:  "Admin",
+		Email: "johndoe@example.org",
+	}
```
System-Glitch, line 192:
```suggestion
	user := &TestUser{}
```

System-Glitch (changes requested):
Implementation look good to me! 👍🏻 
I found a few issues with the tests that need fixing.

auth/jwt_controller.go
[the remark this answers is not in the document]
Morishiri: I would keep both and only remove it when there is new major release. Also I would go back to the implementation I proposed initially (not allow user to override the mandatory claims) or mention that `sub` is required.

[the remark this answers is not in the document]
System-Glitch: Let's keep `userid` for now. And we will replace it with `sub` in the next major version. I'd like to keep the possibility for the developer to override claims if he wants to, that adds some flexibility. For example, developers may want to generate the `nbf` and `exp` claims dynamically.

auth/jwt_test.go
```diff
@@ -23,14 +24,6 @@ func (suite *JWTAuthenticatorTestSuite) SetupSuite() {
 	database.Migrate()
 }
 
-func (suite *JWTAuthenticatorTestSuite) SetupTest() {
-	user := &TestUser{
-		Name:  "Admin",
-		Email: "johndoe@example.org",
-	}
-	database.GetConnection().Create(user)
-}
```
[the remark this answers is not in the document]
Morishiri: It will not be conflicting, becuse we have `TearDownTest` with `suite.ClearDatabase()` called.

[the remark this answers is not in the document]
System-Glitch: Oh that's right, I didn't see that!

auth/jwt_test.go
```diff
@@ -13,6 +14,7 @@ import (
 
 type JWTAuthenticatorTestSuite struct {
 	goyave.TestSuite
+	user *TestUser
```
Morishiri, line 17: @System-Glitch could you help me with this issues? what is the linter complaining about exactly?

System-Glitch: This is a new linter I enabled yesterday. This linter ensure all structs have minimal padding so they are taking less space. (Learn more [here](https://dave.cheney.net/2015/10/09/padding-is-hard)

Just move that field above `goyave.TestSuite` and that warning will go away.

System-Glitch (approved):
Looks goot to me! 👍🏻 
Once the lint issues are fixed I'm ready to merge.

System-Glitch:
Merged! :tada: Thank you again for your contribution!

Issue #137, opened 2021-03-26
Asymmetric JWT signatures support


System-Glitch:
## Proposal

Currently, the built-in JWT is using `jwt.SigningMethodHS256`. Using another method is not possible without re-implementing entirely the `GenerateToken` function and the `JWTController`. Adding support for customizable signing method, and more specifically, supporting asymmetric signing methods. That would enable federated or central authentication without leaking the JWT secret, which is an obvious security concern.

Implementation details can be discussed below.

### Possible drawbacks

`None`.

PR #135, opened 2021-03-18
Add a helper map useful for constructing a json response

agbaraka:
## Description

I have added an alias to map[string]interface{} in the helper package.

This is useful when you want to create a long nested json response in the response.JSON function

If your pull request is functional (new utilities, helpers or middleware for example), **include some usage examples**.
For example:

Normally you would write the response as follows:
``` golang
response.JSON(200, map[string]interface{}{
		"name": "Albert Shirima",
		"website": "albertshirima.com",
		"projects": []map[string]interface{}{
			{
				"name": "Goyave",
				"is_contributor": true,
				"meta": map[string]interface{}{
					"website": "goyave.dev",
					"contributors": 2,
				},
			},
		},
	})

```

With the helper struct:
``` golang
response.JSON(200, helper.Map{
    "name": "Albert Shirima",
    "website": "albertshirima.com",
    "projects": []helper.Map{
	    {
		    "name": "Goyave",
		    "is_contributor": true,
		    "meta": helper.Map{
			    "website": "goyave.dev",
			    "contributors": 2,
		    },
	    },
    },
})
```

### Possible drawbacks

None

## Related issue(s)

List all related issues here:

None

Issue #42, opened 2020-03-06
OpenAPI generator


System-Glitch:
## Proposal

Develop a module able to generate OpenAPI specification for Goyave applications by reading the main route registrer and validation rule sets.

This feature could be included in the CLI Utility ( #39 )

### Possible drawbacks

`None`.

## Additional information

This issue is a feature proposal and is meant to be discussed.
It is also a good candidate if you want to contribute to the project.
This should probably be developed as a side-project.

System-Glitch:
I had a few ideas to make this feature:
- Make it build-in (it would be implemented in a new openapi.go file at the root of the project, to be able to access unexported fields)
- A lot of information can be retrieved simply by reading the route definition and its validation rules.
- The route description can be retrieved using handler comments and godoc.
- The hardest part is not mandatory: getting information about the responses.
- Adding an optional (nullable) struct to the `Route` struct, containing OpenAPI information that can be added manually, such as potential responses, examples, etc.
- For ease of use, it could be integrated into the CLI utility that is being developed.

njr8392:
I'd love to help you with this if you have't already started.

System-Glitch:
Hello @njr8392 ! Thank you very much for your help ! Don't hesitate to contact me (via email, telegram or discord) if you have any question or if you need guidance.

njr8392:
appreciate it!I agree that reading the routes/validation requirements would do the bulk of the job or using godoc to parse comments.  How would you ideally like this implemented? Should i assume that the user will declare routes in a routes.go file? Maybe read all files to look for route declaration/validation and parse that info and put into a document format?  Is your goal solely to produce automatically generated documentationfor the api the user creates?Thanks, NickSent from my Verizon, Samsung Galaxy smartphone
-------- Original message --------From: SystemGlitch <notifications@github.com> Date: 10/22/20  2:56 AM  (GMT-05:00) To: System-Glitch/goyave <goyave@noreply.github.com> Cc: Nicholas Rodgers <nrodgers88@gmail.com>, Mention <mention@noreply.github.com> Subject: Re: [System-Glitch/goyave] OpenAPI generator (#42) 
Hello @njr8392 ! Thank you very much for your help ! Don't hesitate to contact me (via email, telegram or discord) if you have any question or if you need guidance.

System-Glitch:
[...]

PR #141, opened 2021-04-28
v3.8.0

System-Glitch:
## Description

- Added `helper.Map`. *Thanks to @agbaraka  for the contribution!* 
- Added `router.Group`, which is an alias for `router.Subrouter("")`, aimed at improving route definition readability.
- Address generation outputs `127.0.0.1` if `0.0.0.0` is set as server host.
- Added validation `Rule.IsType()` and `Rule.IsTypeDependent()`.
- Added `route.BuildURI()` to be able to generate a full URI without the host and protocol prefix.
- Added various accessors for routes and routers.
- Fixed timing attack on config basic authenticator.
- Exported `goyave.NewRouter()`.
- Minor memory improvements by re-aligning some structure fields.
- Added support for custom JWT claims. (`jwt.GenerateTokenWithClaims()`). *Thanks to @Morishiri  for the contribution!* 
- Added `JWTController.TokenFunc` to let developers define their own token generation logic. *Thanks to @Morishiri  for the contribution!*
- Added support for RSA and ECDSA JWT signatures.
- Added `JWTAuthenticator.ClaimName`. This field defines what is the name if the claim used as an ID.
- `JWTAuthenticator` now adds token claims to `request.Extra` if the provided token is valid. (key: `"jwt_claims"`)
- `JWTController.Login` now returns `401 Unauthorized` instead of `422 Unprocessable Entity` if provided credentials are invalid.

### Possible drawbacks

`None`.

## Related issue(s)

* #42 
* #135
* #137 
* #138
* #139 
* #140
* #142


goyave_test.go
```diff
@@ -60,6 +60,13 @@ func (suite *GoyaveTestSuite) TestGetAddress() {
 	suite.Equal("https://test.system-glitch.me", getAddress("https"))
 
 	suite.Equal(getAddress("http"), BaseURL())
+
+	config.Set("server.domain", "")
+	config.Set("server.host", "0.0.0.0")
+	config.Set("server.port", 1235.0)
```
ethicnology, line 66: Shouldn't we use integers instead of floats for port?

ethicnology (approved):

goyave_test.go
System-Glitch: It is converted to integer automatically because the type of this configuration entry is set to `reflect.Int`. This test was written before the v3 configuration revamp.

ethicnology: Ok i've pushed a commit which change these values #142

PR #142, opened 2021-04-28
floats to integers

ethicnology:
## Description

This commit change floats values to integers in goyave_test.go as we discussed in the #141 review
```go
// BEFORE
config.Set("server.port", 80.0)
// AFTER
config.Set("server.port", 80)
```


System-Glitch (approved):

System-Glitch:
Thank you! 🎉

PR #162, opened 2021-08-19
Remove name parameter in Cookies method

agbaraka:
The name parameter is unused inside the method

## Description

Name parameter in the Cookies method signature is unused.

### Possible drawbacks

Introduces breaking changes


System-Glitch:
Thank you for this! I'll merge when the test is fixed.

System-Glitch:
🎉 Merged, thank you!

PR #192, opened 2023-04-03
fixed interface compatibility issues with testify v1.8.2

imuni4fun:
## Description

This PR updates the `ITestSuite interface` to be compatible with the next version of testify `v1.8.2` which adds a method to its `TestingSuite interface` which this mimics. Without this change, packages that depend on the newer testify library will cause a conflict at build time.

### Possible drawbacks

None. Since this interface is used internally, it should not impact other packages that depend on either testify version.

## Related issue(s)

None

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,7 +8,7 @@ require (
 	github.com/google/uuid v1.3.0
 	github.com/gorilla/websocket v1.5.0
 	github.com/imdario/mergo v0.3.13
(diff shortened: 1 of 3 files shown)


System-Glitch:
Thank you, this is a change that was done on the v5 branch already but I forgot to apply it to master.

I will release a patch version tomorrow.

imuni4fun:
Awesome! Thanks for the quick turn around also!

Issue #202, opened 2024-06-11
New validator: DoesntStartWith


System-Glitch:
Add a [new validator](https://goyave.dev/basics/validation.html#custom-validators) that checks if all the string under validation doesn't start with the given parameter. You can use the validators defined in `validation/string.go` as a reference. Add the necessary tests, documentation and default language entries.

PR #207, opened 2024-06-14
Add DoesntStartWithValidator

saltyFamiliar:
## References

**Issue(s):** #202   

## Description

I implemented `DoesntStartWithValidator` using the validators in `validation/string.go` as a reference. I also added a corresponding test in `validation/string_test.go` and updated `lang/default.go` with two entries: `"doesnt_start_with"` and `"doesnt_start_with.element"`.

### Possible drawbacks

`DoesntStartWithValidator.Validate` returns `true` for non-string values. Maybe it should return some kind of error.


validation/string.go
```diff
@@ -91,3 +91,36 @@ func (v *EndsWithValidator) MessagePlaceholders(_ *Context) []string {
...
+// with any of the specified prefixes.
+type DoesntStartWithValidator struct {
+	BaseValidator
+	Prefix []string
+}
+
+// Validate checks the field under validation satisfies this validator's criteria.
+func (v *DoesntStartWithValidator) Validate(ctx *Context) bool {
+	val, ok := ctx.Value.(string)
+	return !ok || !lo.ContainsBy(v.Prefix, func(prefix string) bool {
```
System-Glitch, line 107: To keep it consistent with `starts_with` and `ends_with`, the validator should return `false` if the value is not a `string`.

validation/string_test.go
```diff
@@ -126,3 +126,45 @@ func TestEndsWithValidator(t *testing.T) {
...
+		{value: "", want: true},
+		{value: 'a', want: true},
+		{value: 2, want: true},
+		{value: 2.5, want: true},
+		{value: map[string]any{"a": 1}, want: true},
+		{value: []string{"string"}, want: true},
+		{value: []string{"pre-string"}, want: true},
+		{value: []string{"fix-string"}, want: true},
+		{value: true, want: true},
+		{value: nil, want: true},
```
System-Glitch, line 158: Related to my previous comment, all of these cases should expect `false`.

System-Glitch (changes requested):

System-Glitch (approved):

System-Glitch:
Thank you very much! 🎉

Issue #201, opened 2024-06-11
New validator: KeysIn


System-Glitch:
Add a [new validator](https://goyave.dev/basics/validation.html#custom-validators) that checks if all the keys in the object under validation are part of the parameter slice:
```go
v.KeysIn("key1", "key2")
```

The following objects would pass:
```json
{
    "key1": 123
}
{
    "key1": 123,
    "key2": 123
}
```

The following objects would NOT pass:
```json
{
    "key3": 123
}
{
    "key1": 123,
    "key3": 123
}
{
    "key1": 123,
    "key2": 123,
    "key3": 123
}
```

PR #208, opened 2024-06-17
feat: Add KeysInValidator for validating fields containing specific keys

bangadam:
## References
Issue(s): #201 
Discussion: None

## Description
This validator ensures that all keys in the object under validation are part of the specified parameter slice. The main benefits of this addition are:
- It provides a simple and efficient way to validate that an object contains only the allowed keys, which enhances data integrity and consistency.
- It reduces the need for custom validation logic in individual projects, promoting code reusability and maintainability.

### Possible drawbacks
- The KeysIn validator assumes that the object being validated is a map[string]interface{}. If the context value is not of this type, the validation will fail


validation/keysin_test.go
```diff
@@ -0,0 +1,72 @@
+package validation
+
+import "testing"
+
+func TestKeysInValidator_Validate(t *testing.T) {
```
System-Glitch, line 5: Please add a test for the validator constructor (checking it's `Name()`, `IsType()`, `IsTypeDependent()` and `MessagePlaceholders()`.

validation/keysin_test.go
```diff
@@ -0,0 +1,72 @@
...
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			v := &KeysInValidator{
+				BaseValidator: tt.fields.BaseValidator,
+				Keys:          tt.fields.Keys,
+			}
+			if got := v.Validate(tt.args.ctx); got != tt.want {
+				t.Errorf("KeysInValidator.Validate() = %v, want %v", got, tt.want)
```
System-Glitch, line 68: You can use `testify` for assertions.

validation/keysin.go
```diff
@@ -0,0 +1,39 @@
...
+	obj, ok := ctx.Value.(map[string]interface{})
+	if !ok {
+		return false
+	}
+
+	for _, key := range v.Keys {
+		if _, ok := obj[key]; !ok {
+			return false
+		}
+	}
```
System-Glitch, line 22: This validator should check that the object doesn't contain a key that is not allowed. Here, you are checking that the object contains all the keys, which is different.

System-Glitch (changes requested):
You forgot to add the default language lines in `lang/default.go`.

System-Glitch:
> The KeysIn validator assumes that the object being validated is a map[string]interface{}. If the context value is not of this type, the validation will fail

It is correct and this is how it should be done. 👍🏻

System-Glitch:
Implementation looks correct to me. The third test case should expect `true` instead of `false`.

bangadam:
> Implementation looks correct to me. The third test case should expect `true` instead of `false`.

updated, please check

validation/string.go
```diff
@@ -124,3 +124,46 @@ func (v *DoesntStartWithValidator) MessagePlaceholders(_ *Context) []string {
 func DoesntStartWith(prefix ...string) *DoesntStartWithValidator {
 	return &DoesntStartWithValidator{Prefix: prefix}
 }
+
+//------------------------------
+
+// KeysInValidator validates the field under validation must only contain the given keys.
```
System-Glitch, line 130:
```suggestion
// KeysInValidator the field under validation must be an object and all its keys must
// be equal to one of the given values.
```

validation/string.go
```diff
@@ -124,3 +124,46 @@ func (v *DoesntStartWithValidator) MessagePlaceholders(_ *Context) []string {
...
+func (v *KeysInValidator) Name() string { return "keys_in" }
+
+// MessagePlaceholders returns the ":values" placeholder.
+func (v *KeysInValidator) MessagePlaceholders(_ *Context) []string {
+	return []string{
+		":values", strings.Join(v.Keys, ", "),
+	}
+}
+
+// KeysIn creates a new KeysInValidator.
```
System-Glitch, line 166:
```suggestion
// KeysIn the field under validation must be an object and all its keys must
// be equal to one of the given values.
```

validation/string_test.go
```diff
@@ -168,3 +168,34 @@ func TestDoesntStartWithValidator(t *testing.T) {
...
+		assert.False(t, v.IsType())
+		assert.False(t, v.IsTypeDependent())
+		assert.Equal(t, []string{":values", "a, b, c"}, v.MessagePlaceholders(&Context{}))
+	})
+
+	cases := []struct {
+		value []string
+		input any
+		want  bool
+	}{
```
System-Glitch, line 186:
```suggestion
	cases := []struct {
		input any
		value []string
		want  bool
	}{
```

validation/string_test.go
```diff
@@ -168,3 +168,34 @@ func TestDoesntStartWithValidator(t *testing.T) {
...
+		{value: []string{"a", "b", "c"}, input: map[string]any{"a": 1, "b": 2, "c": 3}, want: true},
+		{value: []string{"a", "b", "c"}, input: map[string]any{"a": 1, "b": 2, "c": 3, "d": 4}, want: false},
+		{value: []string{"a", "b", "c"}, input: map[string]any{"a": 1, "b": 2}, want: true},
+	}
+
+	for _, tc := range cases {
+		tc := tc
+		t.Run(fmt.Sprintf("Validate_%v_%t", tc.value, tc.want), func(t *testing.T) {
```
System-Glitch, line 194:
```suggestion
		t.Run(fmt.Sprintf("Validate_%v_%t", tc.input, tc.want), func(t *testing.T) {
```

validation/string_test.go
```diff
@@ -168,3 +168,34 @@ func TestDoesntStartWithValidator(t *testing.T) {
 		})
 	}
 }
+
+func TestKeysInValidator(t *testing.T) {
+	t.Run("Constructor", func(t *testing.T) {
+		v := KeysIn([]string{"a", "b", "c"}...)
```
System-Glitch, line 174:
```suggestion
		v := KeysIn("a", "b", "c")
```

validation/string_test.go
```diff
@@ -168,3 +168,34 @@ func TestDoesntStartWithValidator(t *testing.T) {
...
+	})
+
+	cases := []struct {
+		value []string
+		input any
+		want  bool
+	}{
+		{value: []string{"a", "b", "c"}, input: map[string]any{"a": 1, "b": 2, "c": 3}, want: true},
+		{value: []string{"a", "b", "c"}, input: map[string]any{"a": 1, "b": 2, "c": 3, "d": 4}, want: false},
+		{value: []string{"a", "b", "c"}, input: map[string]any{"a": 1, "b": 2}, want: true},
```
System-Glitch, line 189:
```suggestion
		{value: []string{"a", "b", "c"}, input: map[string]any{"a": 1, "b": 2, "c": 3}, want: true},
		{value: []string{"a", "b", "c"}, input: map[string]any{"a": 1, "b": 2, "c": 3, "d": 4}, want: false},
		{value: []string{"a", "b", "c"}, input: map[string]any{"a": 1, "b": 2}, want: true},
		{value: []string{"a", "b", "c"}, input: map[string]any{}, want: true},
		{value: []string{"a", "b", "c"}, input: "", want: false},
		{value: []string{"a", "b", "c"}, input: 'a', want: false},
		{value: []string{"a", "b", "c"}, input: 2, want: false},
		{value: []string{"a", "b", "c"}, input: 2.5, want: false},
		{value: []string{"a", "b", "c"}, input: true, want: false},
		{value: []string{"a", "b", "c"}, input: nil, want: false},
		{value: []string{"a", "b", "c"}, input: (map[string]any)(nil), want: true},
```

System-Glitch (approved):

System-Glitch:
Thank you very much! 🎉

Issue #203, opened 2024-06-11
New validator: DoesntEndWith


System-Glitch:
Add a [new validator](https://goyave.dev/basics/validation.html#custom-validators) that checks if all the string under validation doesn't start with the given parameter. You can use the validators defined in `validation/string.go` as a reference. Add the necessary tests, documentation and default language entries.

PR #209, opened 2024-06-17
chore: Add DoesntEndWithValidator to validate that a string doesn't end with a given suffix

bangadam:
## References

**Issue(s):** #203  
**Discussion:** None

## Description

This validator ensures that the string under validation does not end with the specified suffix. The main benefits of this addition are:

- It provides a simple and efficient way to validate that a string does not end with a given suffix, enhancing data validation and integrity.
- It reduces the need for custom validation logic in individual projects, promoting code reusability and maintainability.

### Possible Drawbacks
- The `DoesntEndWith` validator assumes the validated object is a string. If the context value is not of this type, the validation will fail


validation/string_end.go
```diff
@@ -0,0 +1,36 @@
+package validation
```
System-Glitch, line 1: Please move it to `validation/string.go`.

validation/string_end.go
```diff
@@ -0,0 +1,36 @@
...
+}
+
+// Validate checks if the field under validation satisfies this validator's criteria.
+func (v *DoesntEndWithValidator) Validate(ctx *Context) bool {
+	str, ok := ctx.Value.(string)
+	if !ok {
+		return false
+	}
+
+	return !strings.HasSuffix(str, v.Suffix)
```
System-Glitch, line 20: To make it consistent with other prefix/suffix validators, `v.Suffix` should be a `[]string` and the validator should check that the value doesn't end with **any** of the suffixes.

System-Glitch (changes requested):
You forgot to add the default language lines in `lang/default.go`.

System-Glitch (approved):

System-Glitch:
Thank you very much! 🎉

Issue #210, opened 2024-06-20
Validation of config struct and slice


BowlOfSoup:
This is part question, part proposal. First off, thanks for releasing v5 for Goyave. I saw the project on Reddit and I started to rewrite an app built with loosely coupled packages into Goyave.

I've added my own custom configuration, for example:

```json
"some_config_list_of_endpoints": [
    {
        "name": "dummy entry",
        "api_key": "value123",
        "url": "https://example.com",
    },
    {
        "name": "yet another entry",
        "api_key": "uuid-something-something",
        "url": "https://some-url.com",
    }
],
"custom_auth_service": {
    "url": "http://localhost-for-testing:1234",
    "realm": "development",
    "keys_refresh_minutes": 15,
    "verify_with_issuer": false
}
```

The values in the object `custom_auth_service` are validated successfully. I'm defining that part like: 

```go
config.Register("custom_auth_service.url", config.Entry{
	Value: "",
	Type:  reflect.String,
})

config.Register("custom_auth_service.keys_refresh_minutes", config.Entry{
	Value: "",
	Type:  reflect.Int,
})

// etc.
```

**Question 1**; I would like all values to be mandatory. How can I achieve that? Should I build a custom validator for that?

**Question 2**; I would like my config to be parsed into a struct instead of having to do `server.Config().Get("custom_auth_service.keys_refresh_minutes")` somewhere, I would like to fetch `custom_auth_service` directly and get a struct. Is this possible? (I know currently it's not according to documentation and trial)

Then I've created a slice of configs, the `some_config_list_of_endpoints` category. 

**Question 3**; Does Goyave support this configuration? I can fetch the entries by doing `server.Config().Get("some_config_list_of_endpoints").([]interface{})` but I can't seem to be able to 'init' or register the configuration without getting errors:

```go
config.Register("some_config_list_of_endpoints", config.Entry{
	Value:   []ListOfEndpoints{},
	Type:    reflect.Slice,
	IsSlice: true,
})

config.Register("some_config_list_of_endpoints.name", config.Entry{
	Value: "",
	Type:  reflect.String,
})
```

This will give me the error: `panic: attempted to add an entry to non-category "some_config_list_of_endpoints"` (on the `.name` `Register()`.

The proposal part would be to:
1. Support mandatory values, as in, values can not be empty or nil
2. Be able to parse the config into a defined struct, as in, unmarshall the config JSON

Thanks in advance :)

System-Glitch:
[...]

PR #211, opened 2024-06-20
ISSUE-210 - Add Required property to indicate a config registration c…

BowlOfSoup:
Add Required property to indicate a config registration can't be omitted or nil (+ tests)

## References

**Issue(s):** #210 

## Description
Changes in this PR make it possible to, when registering a config value, to indicate that the value can not be `nil` and also validates omitting the complete config line. Both will trigger a validation message.

### Possible drawbacks/suggestions
Imo we should provide a 'constructor' like `config.NewEntry()` in which we can set default values to prevent this change to the `Entry` struct to be a breaking change for people not using keyed initialization on structs. _This was discussed and marked as not breaking_.

Proposal for constructor `NewEntry()`:

```go
package config

type EntryOpts struct {
    AuthorizedValues []any
    IsSlice bool
    Required bool
}

func NewEntry(value any, type reflect.Kind, opts EntryOpts) *Entry {
    return &Entry{
        Value: "default value",
	Type: reflect.String,
	IsSlice: opts.IsSlice,
	AuthorizedValues: opts.AutorizedValues,
	Required: opts.Required
    }
}

// use it like this
config.Register("customCategory.customEntry", config.NewEntry(
    Value: "default value",
    Type: reflect.String,
    EntryOpts{
        AuthorizedValues: []any{},
        IsSlice: false,
        Required: true,
    }
)
```


config/default.go
```diff
@@ -4,49 +4,49 @@ import "reflect"
...
-			"allowGlobalUpdate":                        &Entry{false, []any{}, reflect.Bool, false},
-			"disableAutomaticPing":                     &Entry{false, []any{}, reflect.Bool, false},
-			"disableForeignKeyConstraintWhenMigrating": &Entry{false, []any{}, reflect.Bool, false},
+			"skipDefaultTransaction":                   &Entry{false, []any{}, reflect.Bool, false, false},
+			"dryRun":                                   &Entry{false, []any{}, reflect.Bool, false, false},
+			"prepareStmt":                              &Entry{true, []any{}, reflect.Bool, false, false},
+			"disableNestedTransaction":                 &Entry{false, []any{}, reflect.Bool, false, false},
+			"allowGlobalUpdate":                        &Entry{false, []any{}, reflect.Bool, false, false},
+			"disableAutomaticPing":                     &Entry{false, []any{}, reflect.Bool, false, false},
+			"disableForeignKeyConstraintWhenMigrating": &Entry{false, []any{}, reflect.Bool, false, false},
```
System-Glitch, line 49: All of them except `server.proxy.host` should be required.

The validation is also performed when calling `config.Set()`. It is incorrect to set them to `nil` so it shouldn't be allowed.

config/entry.go
```diff
@@ -29,17 +30,22 @@ func makeEntryFromValue(value any) *Entry {
 		kind = t.Elem().Kind()
 		isSlice = true
 	}
-	return &Entry{value, []any{}, kind, isSlice}
+	return &Entry{value, []any{}, kind, isSlice, false}
 }
 
 func (e *Entry) validate(key string) error {
-	if e.Value == nil { // nil values means unset
+	if e.Value == nil {
```
System-Glitch, line 37: A simple `nil` comparison may silently fail in some scenarios. For example if the concrete type of `e.Value` is not `any`.

[Playground](https://goplay.tools/snippet/bBFLhCN0DLi)

We should check the reflect value instead:
```go
v := reflect.ValueOf(strct.Value)
ok := v.IsValid() && !v.IsNil()
```

To avoid multiplying reflect calls, let's replace the `reflect.TypeOf(e.Value)` with `reflect.ValueOf(e.Value)` and do the check just before `tryConversion()`. And remove `list := reflect.ValueOf(e.Value)` below.

config/entry.go
```diff
@@ -230,3 +202,40 @@ func (e *Entry) convertEnvVar(str, key string) (any, error) {
...
+	if e.Required {
+		switch kind {
+		case reflect.String:
+			if v.Len() == 0 {
+				return errors.Errorf("%q is required and cannot be an empty string", key)
+			}
+		case reflect.Map:
+			if len(v.MapKeys()) == 0 {
+				return errors.Errorf("%q is required and cannot be an empty map", key)
+			}
```
System-Glitch, line 218: Maps are not supported. They cannot be told apart from categories.

config/entry.go
```diff
@@ -192,3 +202,40 @@ func (e *Entry) convertEnvVar(str, key string) (any, error) {
 
 	return nil, nil
 }
+
+func (e *Entry) validateRequired(kind reflect.Kind, key string) error {
```
System-Glitch, line 206: I think we should only check is the value is `nil`. There are some legitimate use-cases where you would want to use `0` or an empty string as a value.

The main benefit of adding a required flag is to notify developers when they forgot to add an entry in their config file. We should limit ourselves to this in my opinion.

config/entry.go
```diff
@@ -29,17 +30,22 @@ func makeEntryFromValue(value any) *Entry {
...
 	}
-	return &Entry{value, []any{}, kind, isSlice}
+	return &Entry{value, []any{}, kind, isSlice, false}
 }
 
 func (e *Entry) validate(key string) error {
-	if e.Value == nil { // nil values means unset
+	if e.Value == nil {
+		if e.Required {
+			return errors.Errorf("%q is required and must contain a value", key)
```
System-Glitch, line 39:
```suggestion
			return errors.Errorf("%q is required", key)
```

System-Glitch (changes requested):
This doesn't look like a breaking change to me if we drop the "empty" validation. It would only be breaking for codebases not always using keyed struct fields, which is very rare and usually not recommended anyway. The behavior for entries without `Required: true` stays the same as before.

config/entry.go
BowlOfSoup: Only checking if the value is `nil` won't cover someone making **incorrect** configuration, for example in the case of containerizing a project, where (e.g. with Kubernetes) env variables are injected directly in the container and then used in the config. 

If someone explicitly wants to use empty or 0, they can choose not to make it 'required'. 

Then again, I would rather like support for the validation RuleSet to be supported for the configuration so devs can decide for themselves what validation they want. What is your opinion on that?

config/default.go
BowlOfSoup: > This doesn't look like a breaking change to me if we drop the "empty" validation. It would only be breaking for codebases not always using keyed struct fields, which is very rare and usually not recommended anyway. 

Agreed. **But**, we can't dictate how people use the framework; There is risk if a new (minor or patch) release is done with a change to the contract for the `Entry`. It's also done without keyed fields in `config.default.go:73`. 

Your suggestion is to remove the 'empty', but what if in the future more keys need to be added? :)

BowlOfSoup: > All of them except `server.proxy.host` should be required.

Allright. But what if I don't want to set `"maxUploadSize": 10.0,` because I will never upload files and I would be fine with a default value of 10.0?

System-Glitch: > we can't dictate how people use the framework

Actually, we do since the framework is advertised as opinionated. Adding a field to a struct has never been considered a breaking change in any library I have contributed to.

> Your suggestion is to remove the 'empty', but what if in the future more keys need to be added? :)

I'm not sure I understand what you mean. Are you saying that we should think about potentially adding more struct fields to `Entry`?

> But what if I don't want to set "maxUploadSize": 10.0, because I will never upload files and I would be fine with a default value of 10.0?

The default value is 10 so it's not `nil`. Even if you don't provide it in your config file, the validation will not fail.

There are two scenarios:
- Default value is set to `nil` and entry is required: the developer **must** add it to their config file
- Default value is set to something and entry is required: the developer can, but is not forced to add it to their config file. However, they cannot set their value to `nil` (either from the config file or from `config.Set()`.

config/entry.go
BowlOfSoup: I did some tests and you can indeed omit the complete value from the config. Having Required without any additional checks (empty or `nil`) would already benefit indeed.

Let me know what your ideas are on a clean way the handle adding a RuleSet to the config. Are you going to do that in an issue, or in Discord?

config/default.go
BowlOfSoup: > I'm not sure I understand what you mean. Are you saying that we should think about potentially adding more struct fields to `Entry`?

Not right now, but in general. One might add more fields to a struct. _Personally_ I think adding a constructor to a large struct is good in the long term. But... I'm not as experienced as you in Go so, if it's not considered a breaking change: good!

> The default value is 10 so it's not `nil`. Even if you don't provide it in your config file, the validation will not fail.

👍

BowlOfSoup: > All of them except `server.proxy.host` should be required.

Done ✅

config/entry.go
BowlOfSoup: Done ✅ 
A bit different, but check it out 🙂

config/entry.go
BowlOfSoup: Functionality removed ✔️

config/entry.go
```diff
@@ -34,19 +34,19 @@ func makeEntryFromValue(value any) *Entry {
...
 
+	v := reflect.ValueOf(e.Value)
+	if e.Required && (!v.IsValid() || e.Value == nil) {
+		return errors.Errorf("%q is required", key)
+	}
+
 	t := reflect.TypeOf(e.Value)
+	if t == nil {
+		return nil // Can't determine
+	}
```
System-Glitch, line 49:
```suggestion
	t := v.Type()
```

config/entry.go
```diff
@@ -34,19 +34,19 @@ func makeEntryFromValue(value any) *Entry {
...
-		// When just nil, we don't need to validate.
-		return nil
-	}
-
 	if err := e.tryEnvVarConversion(key); err != nil {
 		return err
 	}
 
+	v := reflect.ValueOf(e.Value)
+	if e.Required && (!v.IsValid() || e.Value == nil) {
```
System-Glitch, line 42:
```suggestion
	if e.Required && (!v.IsValid() || v.IsNil()) {
```

config/config_test.go
```diff
@@ -701,7 +708,66 @@ func TestConfig(t *testing.T) {
...
 	})
 }
+
+func TestRequiredConfig(t *testing.T) {
+	defaultLoader.mu.Lock()
+	loader := loader{
+		defaults: make(object, len(defaultLoader.defaults)),
+	}
+	loadDefaults(defaultLoader.defaults, loader.defaults)
+	defaultLoader.mu.Unlock()
```
System-Glitch, line 723:
```suggestion
	loader := loader{
		defaults: make(object),
	}
```
No need to rely on the default loader here. You are not using those entries.

System-Glitch (changes requested):
Very minor adjustments to make, otherwise LGTM! 👍🏻

config/entry.go
BowlOfSoup: This won't work because

```
panic: reflect: call of reflect.Value.Type on zero Value
```

This happens when asserting the config `server.proxy.host`. Which is `nil` and also not required. Imo it's fine if we don't continue with such a value.

config/entry.go
System-Glitch: You're right. Here is what I suggest to cover the scenario I showed in the playground link earlier.
```suggestion
	if e.Required && (!v.IsValid() || e.Value == nil || (v.Kind() == reflect.Pointer && v.IsNil())) {
```

config/entry.go
System-Glitch: It makes sense indeed. No need to validate in that case.

config/entry.go
System-Glitch: I created a [Github discussion](https://github.com/go-goyave/goyave/discussions/212) on this topic. We'll have to think about it more.

System-Glitch:
Thank you very much! 🎉

PR #214, opened 2024-06-26
#213 - Validation: use default language if no language is given in options

### LINKED ISSUE #213: Validation: use default language if not provided
## Proposal

When using validation manually, not providing the `Language` field in the `validation.Options` will trigger a `panic: runtime error: invalid memory address or nil pointer dereference`.

We should load the built-in default language (`en-US`) if the option is omitted. Or we could use an empty language that would return only the name of the entry.

### Possible drawbacks

Loading the default language for every validation can reduce performance a bit. This would still be very much recommended to use the option. This would need to be properly documented.

BowlOfSoup:
Fetch default language if no language is given when validating.

## References

**Issue:** fixes #213

### DIFF
--- a/validation/validator.go
+++ b/validation/validator.go
@@ -275,6 +275,9 @@ func Validate(options *Options) (*Errors, []error) {
 	if options.Extra == nil {
 		options.Extra = map[any]any{}
 	}
+	if options.Language == nil {
+		options.Language = lang.New().GetDefault()
+	}
 
 	rules := options.Rules.AsRules()
 	for _, field := range rules {
--- a/validation/validator_test.go
+++ b/validation/validator_test.go
@@ -283,6 +283,23 @@ func TestValidate(t *testing.T) {
 			wantData: map[string]any{},
 		},
 		{
+			desc: "nil_delete_from_parent_without_language",
+			options: &Options{
+				Data: map[string]any{"property": nil},
+				Rules: RuleSet{
+					{Path: "property", Rules: List{Required()}},
+				},
+			},
+			wantValidationErrors: &Errors{
+				Fields: FieldsErrors{
+					"property": &Errors{
+						Errors: []string{"The property is required."},
+					},
+				},
+			},
+			wantData: map[string]any{},
+		},
+		{
 			desc: "nil_nullable_dont_delete_from_parent",
 			options: &Options{
 				Data:     map[string]any{"property": nil},


validation/validator_test.go
```diff
@@ -282,6 +282,23 @@ func TestValidate(t *testing.T) {
 			},
 			wantData: map[string]any{},
 		},
+		{
+			desc: "nil_delete_from_parent_without_language",
```
System-Glitch, line 286:
```suggestion
			desc: "no_language_provied_in_options",
```

validation/validator.go
```diff
@@ -275,6 +275,9 @@ func Validate(options *Options) (*Errors, []error) {
 	if options.Extra == nil {
 		options.Extra = map[any]any{}
 	}
+	if options.Language == nil {
+		options.Language = lang.New().GetDefault()
```
System-Glitch, line 279: Please add a comment on the `Language` field in the `Options` structure:

```
// Language used for validation error message translation. If `nil` or omitted, defaults to `en-US`.
```

System-Glitch (approved):
LGTM 👍🏻

BowlOfSoup:
> @BowlOfSoup Is this PR ready?

@System-Glitch no it's not. And I even think now we shouldn't fix this at all:
1. Performance penalty
2. Too easy to abuse the possibility of not passing a language. People may not care.

It's up to you to decline 🙂

System-Glitch:
I will think about it. Thank you for your input.

System-Glitch:
To solve the performance issue, let's go back to using an empty language and return the lang entry name:
```go
if options.Language == nil {
	options.Language = &lang.Language{}
}
```
This will require a slight change in the `lang/language.go` file to make sure a `*Language` doesn't panic if its maps are not initialized.

I think it's a good balance between performance and convenience. We'll just need to properly document what happens if the option is not provided.

BowlOfSoup:
> To solve the performance issue, let's go back to using an empty language and return the lang entry name:
> 
> ```go
> if options.Language == nil {
> 	options.Language = &lang.Language{}
> }
> ```
> 
> This will require a slight change in the `lang/language.go` file to make sure a `*Language` doesn't panic if its maps are not initialized.
> 
> I think it's a good balance between performance and convenience. We'll just need to properly document what happens if the option is not provided.

This would return untranslated strings. Is this the possibility we want to give the user? We could, then input = output.

System-Glitch:
> This would return untranslated strings. Is this the possibility we want to give the user? We could, then input = output.
I prefer this over a nil dereference panic.

Another solution would be to return an error (or panic) with an explicit message saying that this option is mandatory.

I think it would be nice to have someone else's opinion on this.

System-Glitch:
@BowlOfSoup I had an idea that would satisfy both the need for translated lines as well as performance:
- Add a getter for the default language in `lang/default.go`:
```go
func GetDefault() *Language {
	return enUS
}
```
- Use this in `validation/validator.go` if the language option is not provided.

The downside is that custom language entries won't be supported this way since only the builtin default ones are loaded. I think this is acceptable. We just need to document it right to make it clear.

BowlOfSoup:
This would indeed be a good solution!Op 26 jul 2024 om 10:30 heeft SystemGlitch ***@***.***> het volgende geschreven:﻿
@BowlOfSoup I had an idea that would satisfy both the need for translated lines as well as performance:

Add a getter for the default language in lang/default.go:

func GetDefault() *Language {
	return enUS
}

Use this in validation/validator.go if the language option is not provided.

The downside is that custom language entries won't be supported this way since only the builtin default ones are loaded. I think this is acceptable. We just need to document it right to make it clear.

System-Glitch:
I implemented it in this commit: 3e4f090a2c5ac7c02cb17cf47a64d6958f9f5ffa

Thank you very much for your suggestion!

PR #217, opened 2024-07-11
#216 - Add message to request (extra) and handle in status handler

### LINKED ISSUE #216: Respond with message for unclear status codes
## Proposal
When developing my application with Goyave, I ran into a http status code `400`. After looking for some time (too long) and debugging Goyave itself, I found out that I gave the Content-Type header `application/json`, but an **empty** body.

This triggered the following piece of code in `middleware/parse/parse.go`:
```go
bodyBytes := bodyBuf.Bytes()
if strings.HasPrefix(contentType, "application/json") {
    var body any
    if err := json.Unmarshal(bodyBytes, &body); err != nil {
        response.Status(http.StatusBadRequest)
    }  
```

I got a `400` without any context. My suggestion is to add a message to the response body that indicates why a certain response code has been returned.

BowlOfSoup:
## References
**Issue(s):** closes #216  

## Description
As per discusion in Discord, I've added a message when the parser returns a `400` bad request when Content-Type JSON is given, but no, or an invalid body. I've handled this via `response.Extra[goyave.ExtraRequestError{}]`. 

It also translates.

### Possible drawbacks
Let me know :)

### DIFF
--- a/middleware/parse/parse.go
+++ b/middleware/parse/parse.go
@@ -3,7 +3,6 @@ package parse
 import (
 	"bytes"
 	"encoding/json"
-	"fmt"
 	"io"
 	"net/http"
 	"net/url"
@@ -75,7 +74,7 @@ func (m *Middleware) Handle(next goyave.Handler) goyave.Handler {
 					if err := json.Unmarshal(bodyBytes, &body); err != nil {
 						response.Status(http.StatusBadRequest)
 						r.Extra[goyave.ExtraRequestError{}] = []error{
-							fmt.Errorf("request.json-invalid-body"),
+							goyave.ErrInvalidJSONBody,
 						}
 					}
 					r.Data = body
--- a/middleware/parse/parse.go
+++ b/middleware/parse/parse.go
@@ -3,6 +3,7 @@ package parse
 import (
 	"bytes"
 	"encoding/json"
+	"fmt"
 	"io"
 	"net/http"
 	"net/url"
@@ -73,6 +74,9 @@ func (m *Middleware) Handle(next goyave.Handler) goyave.Handler {
 					var body any
 					if err := json.Unmarshal(bodyBytes, &body); err != nil {
 						response.Status(http.StatusBadRequest)
+						r.Extra[goyave.ExtraRequestError{}] = []error{
+							fmt.Errorf("request.json-empty-body"),
+						}
 					}
 					r.Data = body
 				} else {
--- a/middleware/parse/parse.go
+++ b/middleware/parse/parse.go
@@ -75,7 +75,7 @@ func (m *Middleware) Handle(next goyave.Handler) goyave.Handler {
 					if err := json.Unmarshal(bodyBytes, &body); err != nil {
 						response.Status(http.StatusBadRequest)
 						r.Extra[goyave.ExtraRequestError{}] = []error{
-							fmt.Errorf("request.json-empty-body"),
+							fmt.Errorf("request.json-invalid-body"),
 						}
 					}
 					r.Data = body
--- a/middleware/parse/parse.go
+++ b/middleware/parse/parse.go
@@ -47,6 +47,7 @@ func (m *Middleware) Handle(next goyave.Handler) goyave.Handler {
 	return func(response *goyave.Response, r *goyave.Request) {
 		if err := parseQuery(r); err != nil {
 			response.Status(http.StatusBadRequest)
+			r.Extra[goyave.ExtraParseError{}] = fmt.Errorf("%w: %w", goyave.ErrInvalidQuery, err)
 			return
 		}
 
@@ -83,10 +84,12 @@ func (m *Middleware) Handle(next goyave.Handler) goyave.Handler {
 					r.Data, err = generateFlatMap(req, maxSize)
 					if err != nil {
 						response.Status(http.StatusBadRequest)
+						r.Extra[goyave.ExtraParseError{}] = fmt.Errorf("%w: %w", goyave.ErrInvalidContentForType, err)
 					}
 				}
 			} else {
 				response.Status(http.StatusBadRequest)
+				r.Extra[goyave.ExtraParseError{}] = fmt.Errorf("%w: %w", goyave.ErrErrorInRequestBody, err)
 			}
 		}
 
--- a/middleware/parse/parse.go
+++ b/middleware/parse/parse.go
@@ -3,6 +3,7 @@ package parse
 import (
 	"bytes"
 	"encoding/json"
+	"fmt"
 	"io"
 	"net/http"
 	"net/url"
@@ -73,9 +74,7 @@ func (m *Middleware) Handle(next goyave.Handler) goyave.Handler {
 					var body any
 					if err := json.Unmarshal(bodyBytes, &body); err != nil {
 						response.Status(http.StatusBadRequest)
-						r.Extra[goyave.ExtraRequestError{}] = []error{
-							goyave.ErrInvalidJSONBody,
-						}
+						r.Extra[goyave.ExtraParseError{}] = fmt.Errorf("%w: %w", goyave.ErrInvalidJSONBody, err)
 					}
 					r.Data = body
 				} else {
--- a/middleware/parse/parse_test.go
+++ b/middleware/parse/parse_test.go
@@ -192,7 +192,7 @@ func TestParseMiddleware(t *testing.T) {
 		request.Header().Set("Content-Type", writer.FormDataContentType())
 
 		result := server.TestMiddleware(&Middleware{}, request, func(resp *goyave.Response, _ *goyave.Request) {
-			resp.Status(http.StatusBadRequest)
+			resp.Status(http.StatusOK)
 		})
 
 		assert.NoError(t, result.Body.Close())
--- a/middleware/parse/parse_test.go
+++ b/middleware/parse/parse_test.go
@@ -2,7 +2,6 @@ package parse
 
 import (
 	"bytes"
-	"fmt"
 	"mime/multipart"
 	"net/http"
 	"strings"
@@ -109,7 +108,11 @@ func TestParseMiddleware(t *testing.T) {
 		assert.NoError(t, result.Body.Close())
 		assert.Equal(t, http.StatusBadRequest, result.StatusCode)
 		assert.NotNil(t, request.Extra[goyave.ExtraRequestError{}])
-		assert.Contains(t, request.Extra[goyave.ExtraRequestError{}].([]error), fmt.Errorf("request.json-invalid-body"))
+		assert.NotPanics(t, func() {
+			extraErrors, ok := request.Extra[goyave.ExtraRequestError{}].([]error)
+			require.True(t, ok)
+			assert.Contains(t, extraErrors, goyave.ErrInvalidJSONBody)
+		})
 	})
 
 	t.Run("Multipart", func(t *testing.T) {
--- a/middleware/parse/parse_test.go
+++ b/middleware/parse/parse_test.go
@@ -109,7 +109,7 @@ func TestParseMiddleware(t *testing.T) {
 		assert.NoError(t, result.Body.Close())
 		assert.Equal(t, http.StatusBadRequest, result.StatusCode)
 		assert.NotNil(t, request.Extra[goyave.ExtraRequestError{}])
-		assert.Contains(t, request.Extra[goyave.ExtraRequestError{}].([]error), fmt.Errorf("request.json-empty-body"))
+		assert.Contains(t, request.Extra[goyave.ExtraRequestError{}].([]error), fmt.Errorf("request.json-invalid-body"))
 	})
 
 	t.Run("Multipart", func(t *testing.T) {
--- a/middleware/parse/parse_test.go
+++ b/middleware/parse/parse_test.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"mime/multipart"
 	"net/http"
+	"net/http/httptest"
 	"strings"
 	"testing"
 
@@ -56,6 +57,12 @@ func TestParseMiddleware(t *testing.T) {
 		})
 		assert.NoError(t, result.Body.Close())
 		assert.Equal(t, http.StatusBadRequest, result.StatusCode)
+		assert.NotNil(t, request.Extra[goyave.ExtraParseError{}])
+		assert.NotPanics(t, func() {
+			extraError, ok := request.Extra[goyave.ExtraParseError{}].(error)
+			require.True(t, ok)
+			assert.ErrorIs(t, extraError, goyave.ErrInvalidQuery)
+		})
 	})
 
 	t.Run("Entity Too Large", func(t *testing.T) {
@@ -99,22 +106,44 @@ func TestParseMiddleware(t *testing.T) {
 		assert.Equal(t, http.StatusOK, result.StatusCode)
 	})
 
-	t.Run("JSON Invalid", func(t *testing.T) {
-		request := testutil.NewTestRequest(http.MethodPost, "/parse", bytes.NewBuffer([]byte("{\"unclosed\"")))
-		request.Lang = server.Lang.GetDefault()
-		request.Header().Set("Content-Type", "application/json")
+	t.Run("JSON Parsing Tests", func(t *testing.T) {
+		tests := []struct {
+			name     string
+			body     []byte
+			expected int
+		}{
+			{
+				name:     "JSON Invalid",
+				body:     []byte("{\"unclosed\""),
+				expected: http.StatusBadRequest,
+			},
+			{
+				name:     "JSON Empty",
+				body:     []byte(""),
+				expected: http.StatusBadRequest,
+			},
+		}
 
-		result := server.TestMiddleware(&Middleware{MaxUploadSize: 0.01}, request, func(_ *goyave.Response, _ *goyave.Request) {
-			assert.Fail(t, "Middleware should not pass")
-		})
-		assert.NoError(t, result.Body.Close())
-		assert.Equal(t, http.StatusBadRequest, result.StatusCode)
-		assert.NotNil(t, request.Extra[goyave.ExtraParseError{}])
-		assert.NotPanics(t, func() {
-			extraError, ok := request.Extra[goyave.ExtraParseError{}].(error)
-			require.True(t, ok)
-			assert.Error(t, extraError)
-		})
+		for _, tt := range tests {
+			t.Run(tt.name, func(t *testing.T) {
+				request := testutil.NewTestRequest(http.MethodPost, "/parse", bytes.NewBuffer(tt.body))
+				request.Lang = server.Lang.GetDefault()
+				request.Header().Set("Content-Type", "application/json")
+
+				result := server.TestMiddleware(&Middleware{MaxUploadSize: 0.01}, request, func(_ *goyave.Response, _ *goyave.Request) {
+					assert.Fail(t, "Middleware should not pass")
+				})
+
+				assert.NoError(t, result.Body.Close())
+				assert.Equal(t, tt.expected, result.StatusCode)
+				assert.NotNil(t, request.Extra[goyave.ExtraParseError{}])
+				assert.NotPanics(t, func() {
+					extraError, ok := request.Extra[goyave.ExtraParseError{}].(error)
+					require.True(t, ok)
+					assert.ErrorIs(t, extraError, goyave.ErrInvalidJSONBody)
+				})
+			})
+		}
 	})
 
 	t.Run("Multipart", func(t *testing.T) {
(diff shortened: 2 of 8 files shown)


status_handler.go
```diff
@@ -48,6 +49,39 @@ func (*ErrorStatusHandler) Handle(response *Response, _ *Request) {
...
+
+	if request.Lang == nil {
+		lang = h.Lang().GetDefault().Name()
+	} else {
+		lang = request.Lang.Name()
+	}
+
+	if e, ok := request.Extra[ExtraRequestError{}]; ok {
+		switch v := e.(type) {
+		case []error:
```
ccoVeille, line 68: Can't you have a simple error? Here you are only supporting a slice of error

middleware/parse/parse_test.go
```diff
@@ -107,6 +108,8 @@ func TestParseMiddleware(t *testing.T) {
 		})
 		assert.NoError(t, result.Body.Close())
 		assert.Equal(t, http.StatusBadRequest, result.StatusCode)
+		assert.NotNil(t, request.Extra[goyave.ExtraRequestError{}])
+		assert.Contains(t, request.Extra[goyave.ExtraRequestError{}].([]error), fmt.Errorf("request.json-invalid-body"))
```
ccoVeille, line 112: You may face a type assertion panic here

middleware/parse/parse.go
```diff
@@ -73,6 +74,9 @@ func (m *Middleware) Handle(next goyave.Handler) goyave.Handler {
 					var body any
 					if err := json.Unmarshal(bodyBytes, &body); err != nil {
 						response.Status(http.StatusBadRequest)
+						r.Extra[goyave.ExtraRequestError{}] = []error{
+							fmt.Errorf("request.json-invalid-body"),
```
ccoVeille, line 78: You should use a sentinel, but also, you shouldn't use fmt.Errorf if you are not formatting.

So please use errors.New and move it at the package level, so you can catch it with errors.Is

ccoVeille (commented):
I know it's a draft, but here is my review

status_handler_test.go
```diff
@@ -131,3 +132,40 @@ func TestValidationStatusHandler(t *testing.T) {
 
 	assert.Equal(t, "{\"error\":{\"body\":{\"fields\":{\"field\":{\"errors\":[\"The field is required\"]}},\"errors\":[\"The body is required\"]},\"query\":{\"fields\":{\"query\":{\"errors\":[\"The query is required\"]}}}}}\n", string(body))
```
ccoVeille, line 133: Please do not escape quote, use backticks

```suggestion
 	assert.Equal(t, `{"error":{"body":{"fields":{"field":{"errors":["The field is required"]}},"errors\":["The body is required"]},"query":{"fields":{"query":{"errors":[\"The query is required\"]}}}}}
`, string(body))
```

status_handler_test.go
```diff
@@ -131,3 +132,40 @@ func TestValidationStatusHandler(t *testing.T) {
...
+
+	handler.Handle(resp, req)
+
+	res := recorder.Result()
+	body, err := io.ReadAll(res.Body)
+
+	assert.NoError(t, res.Body.Close())
+	require.NoError(t, err)
+
+	assert.Equal(t, "{\"error\":[\"The request Content-Type indicates JSON, but the request body is empty or invalid\"]}\n", string(body))
```
ccoVeille, line 151: Please use backticks


```suggestion
	assert.Equal(t, `{"error":["The request Content-Type indicates JSON, but the request body is empty or invalid"]}
`, string(body))
```

status_handler_test.go
```diff
@@ -131,3 +132,40 @@ func TestValidationStatusHandler(t *testing.T) {
...
```
ccoVeille, line 170:
```suggestion
	assert.Equal(t, `{"error":["The request Content-Type indicates JSON, but the request body is empty or invalid"]}
`, string(body))
```

middleware/parse/parse_test.go
BowlOfSoup: You're right, but that would not be a bad thing. This case is testing these lines in code 
```
r.Extra[goyave.ExtraRequestError{}] = []error{
    fmt.Errorf("request.json-invalid-body"),
}
```
which always returns `[]error{}`. If this case ever panics, it probably should.

ccoVeille: You should test it with NotPanics then

https://pkg.go.dev/github.com/stretchr/testify/assert#NotPanics

middleware/parse/parse.go
System-Glitch: Don't use Goyave error wrapping for this those. Collecting the caller frames isn't useful here. You can use the std errors package.

BowlOfSoup: @System-Glitch I already refactored some. Did you check?

request.go
```diff
@@ -28,9 +29,18 @@ type (
...
 	ExtraQueryValidationError struct{}
+
+	// ExtraRequestError the key used in `Context.Extra` to
+	// store specific request errors.
+	ExtraRequestError struct{}
+)
+
+var (
+	// ErrInvalidJSONBody error for invalid JSON body
+	ErrInvalidJSONBody = errors.New("request.json-invalid-body")
```
System-Glitch, line 40: If we are using errors, we should wrap them with the actual error and use `errors.Is` in the status handler to get the corresponding message. I think it would be better.


Simplified example:
```go
r.Extra[goyave.ExtraParseError{}] = fmt.Errorf("%w: %w", goyave.ErrInvalidJSONBody, err)
```

```go
err := request.Extra[ExtraRequestError{}]
switch {
    case errors.Is(err, ErrInvalidJSONBody):
        message := lang.Get("parse.json-invalid-body")
    //...
}

**Review by System-Glitch — COMMENTED** [review]
This is a good first draft 👍🏻 

We should do the same for url-encoded (generateFlatMap) forms and when `io.CopyN` fails.

File: status_handler.go [hunk 8]
```diff
@@ -40,14 +41,52 @@ type ErrorStatusHandler struct {
...
+// Handle generic error responses.
 func (*ErrorStatusHandler) Handle(response *Response, _ *Request) {
 	message := map[string]string{
 		"error": http.StatusText(response.GetStatus()),
 	}
 	response.JSON(response.GetStatus(), message)
 }
 
+// RequestErrorStatusHandler a generic (error) status handler for requests.
+type RequestErrorStatusHandler struct {
```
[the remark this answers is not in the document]
BowlOfSoup: Renaming it to `BadRequest` specifically will constrain the messages we can put in there. What do you think of a more generic `RequestStatusHandler`?

status_handler.go
```diff
@@ -40,14 +41,52 @@ type ErrorStatusHandler struct {
...
+func (h *RequestErrorStatusHandler) Handle(response *Response, request *Request) {
+	var errorMessages []string
+	var langName string
+
+	if request.Lang == nil {
+		langName = h.Lang().GetDefault().Name()
+	} else {
+		langName = request.Lang.Name()
+	}
+	lang := h.Lang().GetLanguage(langName)
```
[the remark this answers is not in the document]
BowlOfSoup: It can and will in tests. Do you have a way around this?

status_handler.go
```diff
@@ -40,14 +41,52 @@ type ErrorStatusHandler struct {
...
+		case string:
+			errorMessages = append(errorMessages, lang.Get(v))
+		default:
+			errorMessages = append(errorMessages, lang.Get(fmt.Sprintf("%v", v)))
+		}
+	}
+
+	messages := map[string][]string{
+		"error": errorMessages,
+	}
```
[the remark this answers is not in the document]
BowlOfSoup: Only supporting one error will limit us. What if there (future?) _are_ multiple errors? Do we want the user to solve 1 and then run into another one? Let me know :)

status_handler.go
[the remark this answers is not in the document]
System-Glitch: What other kind of errors are you thinking about for this status handler? Do you think it could be used for other status codes than 400?

status_handler.go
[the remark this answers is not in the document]
System-Glitch: You can add `req.Lang = server.Lang.GetDefault()` in `prepareStatusHandlerTest()`.

status_handler.go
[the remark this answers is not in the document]
System-Glitch: It depends if we plan on using this status handler for other codes than 400 or not. In any way, the default status handlers should cover the basic and most common use-cases that are global to the entire application. For anything specific, developers are expected to either respond themselves in the controller or override the status handler.

router.go
```diff
@@ -122,6 +122,7 @@ func NewRouter(server *Server) *Router {
 	for i := 400; i <= 418; i++ {
 		router.StatusHandler(&ErrorStatusHandler{}, i)
 	}
+	router.StatusHandler(&RequestErrorStatusHandler{}, http.StatusBadRequest)
```
System-Glitch, line 125:
```suggestion
	for i := 401; i <= 418; i++ {
		router.StatusHandler(&ErrorStatusHandler{}, i)
	}
	router.StatusHandler(&RequestErrorStatusHandler{}, http.StatusBadRequest)
```

ccoVeille: What about 

from := http.StatusUnauthorized // 401
to := http.StatusTooManyRequests // 429

for i := from; i < to; i++

ccoVeille: What about 

from := http.StatusUnauthorized // 401
to := http.StatusTooManyRequests // 429

for i := from; i < to; i++

status_handler.go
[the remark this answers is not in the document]
BowlOfSoup: Yes several. I would like to believe that users of Goyave can also use this for their own handling. Not only 400, but 406, 412, 417, 422 etc. Maybe we don't implement it, but others may. Then this can be used without them having to built their own struct/handler.

I'm fine with `RequestStatusHandler`?

status_handler.go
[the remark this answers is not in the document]
BowlOfSoup: I've removed the possibility now.

request.go
BowlOfSoup: The fist part yes, but why the second part? Isn't that limiting ourselves?

router.go
BowlOfSoup: Renamed all.

middleware/parse/parse_test.go
```diff
@@ -100,13 +101,20 @@ func TestParseMiddleware(t *testing.T) {
...
 		request.Header().Set("Content-Type", "application/json")
 
 		result := server.TestMiddleware(&Middleware{MaxUploadSize: 0.01}, request, func(_ *goyave.Response, _ *goyave.Request) {
 			assert.Fail(t, "Middleware should not pass")
 		})
 		assert.NoError(t, result.Body.Close())
 		assert.Equal(t, http.StatusBadRequest, result.StatusCode)
+		assert.NotNil(t, request.Extra[goyave.ExtraParseError{}])
+		assert.NotPanics(t, func() {
+			extraError, ok := request.Extra[goyave.ExtraParseError{}].(error)
```
ccoVeille, line 114: Your type assertion  is no longer needed


```suggestion
			assert.Error(the, request.Extra[goyave.ExtraParseError{}])
```


You should use


```suggestion
			require.Contains(t, request.Extra,goyave.ExtraParseError{})
     assert.Error(the, request.Extra[goyave.ExtraParseError{}])
```

https://pkg.go.dev/github.com/stretchr/testify/require#Contains

Then the notpanic is no longer needed I think

router.go
```diff
@@ -119,15 +119,19 @@ func NewRouter(server *Server) *Router {
...
 	router.StatusHandler(&ValidationStatusHandler{}, http.StatusUnprocessableEntity)
-	for i := 423; i <= 426; i++ {
+	for i := http.StatusLocked; i <= http.StatusUpgradeRequired; i++ {
 		router.StatusHandler(&ErrorStatusHandler{}, i)
 	}
-	router.StatusHandler(&ErrorStatusHandler{}, 421, 428, 429, 431, 444, 451)
-	router.StatusHandler(&ErrorStatusHandler{}, 501, 502, 503, 504, 505, 506, 507, 508, 510, 511)
+	router.StatusHandler(&ErrorStatusHandler{}, http.StatusMisdirectedRequest, http.StatusPreconditionRequired, http.StatusTooManyRequests, http.StatusRequestHeaderFieldsTooLarge, 444, http.StatusUnavailableForLegalReasons)
```
ccoVeille, line 130: The 444 seems odd.

Is it something special the lib support?

You should add a constant, with a comment, to avoid this magical number here

status_handler.go
```diff
@@ -40,14 +41,42 @@ type ErrorStatusHandler struct {
...
 }
 
+// RequestErrorStatusHandler a generic (error) status handler for requests.
+type RequestErrorStatusHandler struct {
+	Component
+}
+
+// Handle generic request (error) responses.
+func (h *RequestErrorStatusHandler) Handle(response *Response, request *Request) {
+	var errorMessage string
```
ccoVeille, line 59: You should initialize it to "unexpected error"

Otherwise you might have an empty message if you don't enter the if ok

router.go
System-Glitch: `444 No Response` is an unofficial status code used by nginx. It's unlikely that we would return such errors but removing it would be a breaking change.

I appreciate your attention to details and code smells but I don't want to be excessive. I don't mind a few magical numbers here, especially since it's pretty easy to understand that it's a status code.

middleware/parse/parse_test.go
```diff
@@ -100,13 +101,20 @@ func TestParseMiddleware(t *testing.T) {
 
 	t.Run("JSON Invalid", func(t *testing.T) {
 		request := testutil.NewTestRequest(http.MethodPost, "/parse", bytes.NewBuffer([]byte("{\"unclosed\"")))
```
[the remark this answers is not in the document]
BowlOfSoup: I can fix it, but I would really like it if we keep things limited to actual changes in the PR. ;)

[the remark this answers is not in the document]
ccoVeille: I can get it. But when I see oddities. I report them 😁

router.go
ccoVeille: You put more effort in replying me here, than it would be needed to comment the value in the code 😬

I think a constant with a comment would be better. But a comment without constant is at least expected

middleware/parse/parse_test.go
BowlOfSoup: Can't assert an error, if I don't type it as error.

request.go
System-Glitch: I think it is better from a responsibility point of view. We get an error and the status handler renders it. Putting the lang entry as the error message is mixing those concepts together. On top of that, using the real error as a source gives more flexibility to the status handler for rendering it as it contains more details, should we use them one day in the future.

middleware/parse/parse_test.go
ccoVeille: Indeed 🤦‍♀️

status_handler.go
BowlOfSoup: I don't really want to give yet another "unexpected error". I'm hoping to cover all scenarios now :)

request.go
BowlOfSoup: Done.

router.go
BowlOfSoup: I agree with both sides. I've added a constant in response.go (as PHP Symfony also does this).

middleware/parse/parse_test.go
```diff
@@ -146,6 +183,72 @@ func TestParseMiddleware(t *testing.T) {
...
+		request.Lang = server.Lang.GetDefault()
+		request.Header().Set("Content-Type", writer.FormDataContentType())
+
+		result := server.TestMiddleware(&Middleware{}, request, func(resp *goyave.Response, _ *goyave.Request) {
+			resp.Status(http.StatusBadRequest)
+		})
+
+		assert.NoError(t, result.Body.Close())
+		assert.Equal(t, http.StatusBadRequest, result.StatusCode)
+		assert.NotNil(t, request.Extra[goyave.ExtraParseError{}])
```
ccoVeille, line 200: require maybe?

```suggestion
		require.NotNil(t, request.Extra[goyave.ExtraParseError{}])
```

ccoVeille, line 200: This could provide a better error message when it occurs.


```suggestion
		assert.Contains(t, request.Extra, goyave.ExtraParseError{})
```

middleware/parse/parse_test.go
```diff
@@ -146,6 +183,72 @@ func TestParseMiddleware(t *testing.T) {
...
+
+		result := server.TestMiddleware(&Middleware{}, request, func(resp *goyave.Response, _ *goyave.Request) {
+			resp.Status(http.StatusBadRequest)
+		})
+
+		assert.NoError(t, result.Body.Close())
+		assert.NoError(t, resp.Body.Close())
+
+		assert.Equal(t, http.StatusBadRequest, result.StatusCode)
+		assert.NotNil(t, request.Extra[goyave.ExtraParseError{}])
```
ccoVeille, line 243: Same remarks here

router.go
ccoVeille: Thanks for doing the changes, the code is clearer now

middleware/parse/parse_test.go
ccoVeille: Using require should be preferred

BowlOfSoup: `NotNil` is fine for this case.

BowlOfSoup: `assert` is fine. As is `NotNil` and asserts keeps it consistent.

middleware/parse/parse_test.go
BowlOfSoup: See my comments.

status_handler.go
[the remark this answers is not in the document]
System-Glitch: We only check for `ExtraParseError` in this handler so maybe we could name it `ParseErrorStatusHandler`?

System-Glitch (changes requested):

response.go
```diff
@@ -17,6 +17,11 @@ import (
 	"goyave.dev/goyave/v5/util/fsutil"
 )
 
+const (
+	// StatusNginxNoResponse nginx status code: indicates server to return no information to the client and close the connection immediately.
+	StatusNginxNoResponse = 444
+)
+
```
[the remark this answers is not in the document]
ccoVeille: I had suggested it because it made sense to me to have a constant, maybe the variable should either be renamed, or simply removed

request.go
```diff
@@ -28,9 +29,27 @@ type (
...
+	ErrInvalidQuery = errors.New("parse.invalid-query")
+
+	// ErrInvalidJSONBody error when an empty or malformed JSON body is sent.
+	ErrInvalidJSONBody = errors.New("parse.json-invalid-body")
+
+	// ErrInvalidContentForType error when e.g. a multipart form is not actually multipart, or empty.
+	ErrInvalidContentForType = errors.New("parse.invalid-content-for-type")
+
+	// ErrErrorInRequestBody error when e.g. a incoming request is not received properly.
+	ErrErrorInRequestBody = errors.New("parse.error-in-request-body")
```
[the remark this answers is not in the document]
BowlOfSoup: If you have specific tweaking in mind, then I'm fine with merging it like this, since this feels more consistent with other translatable strings. Should be fine do have you do it afterwards to your liking?

middleware/parse/parse_test.go
```diff
@@ -98,15 +106,44 @@ func TestParseMiddleware(t *testing.T) {
...
-		request.Header().Set("Content-Type", "application/json")
+	t.Run("JSON Parsing Tests", func(t *testing.T) {
+		tests := []struct {
+			name     string
+			body     []byte
+			expected int
+		}{
+			{
+				name:     "JSON Invalid",
+				body:     []byte("{\"unclosed\""),
```
ccoVeille, line 117:
```suggestion
				body:     []byte(`{"unclosed"`),
```

BowlOfSoup: Next time we've got to make sure we review it in one go 😛

request.go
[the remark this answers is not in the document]
System-Glitch: Yes don't change anything, I'll see what I do with them after this is merged. Thanks a lot.

middleware/parse/parse_test.go
```diff
@@ -146,6 +183,72 @@ func TestParseMiddleware(t *testing.T) {
...
+			_, err := w.Write([]byte("Partial "))
+			assert.NoError(t, err)
+
+			// Then flush the writer to send the incomplete response
+			if f, ok := w.(http.Flusher); ok {
+				f.Flush()
+			}
+			// Don't close the connection, just stop writing
+		}))
+		defer faultyRequest.Close()
```
System-Glitch, line 223: Couldn't you simply create a `bytes.Buffer` containing this data? This looks convoluted for a simple test. Or you need to simulate a connection that breaks unexpectedly?

middleware/parse/parse_test.go
```diff
@@ -146,6 +183,72 @@ func TestParseMiddleware(t *testing.T) {
...
+		resp, err := client.Get(faultyRequest.URL)
+		require.NoError(t, err)
+
+		// Use the response body from our test server as the request body for our middleware
+		request := testutil.NewTestRequest(http.MethodPost, "/parse", resp.Body)
+		request.Lang = server.Lang.GetDefault()
+		request.Header().Set("Content-Type", "multipart/form-data")
+
+		result := server.TestMiddleware(&Middleware{}, request, func(resp *goyave.Response, _ *goyave.Request) {
+			resp.Status(http.StatusBadRequest)
```
System-Glitch, line 236:
```suggestion
			resp.Status(http.StatusOK)
```
This makes sure the parse middleware doesn't pass because of the response code assertion below. In this case you don't want this line to be executed.

status_handler_test.go
```diff
@@ -129,5 +133,91 @@ func TestValidationStatusHandler(t *testing.T) {
...
+	}
+}
+
+func TestParseErrorStatusHandlerWithNonSupportedErrorStruct(t *testing.T) {
+	req, resp, recorder := prepareStatusHandlerTest()
+
+	handler := &ParseErrorStatusHandler{}
+	handler.Init(resp.server)
+
+	req.Extra[ExtraValidationError{}] = errors.New("some.other.error")
```
System-Glitch, line 209: Remove this line and rename the test with `TestParseErrorStatusHandlerWithoutExtra`.

System-Glitch (approved):
LGTM. I will merge this later with the other pending PRs and release all these changes at once.

middleware/parse/parse_test.go
BowlOfSoup: The last thing yes. It may be able to simplify this test, but... it works for now with exactly the thing I want to test 🙂

middleware/parse/parse_test.go
BowlOfSoup: Done.

middleware/parse/parse_test.go
System-Glitch: Good for me then.

System-Glitch:
Merged. Thank you very much! 🎉

PR #218, opened 2024-07-12
#215 - Response: add support for flushing

### LINKED ISSUE #215: http.Flusher implementation
## Proposal

`*goyave.Response` and built-in chained writers don't implement `http.Flusher`. This makes it difficult for streamed responses for example. Having more options is always a good thing.

We would need to be cautious implementing this in regards to the compression middleware. We'd have to make encoders still work properly when flushing.

**Additional suggestion**: chained writers are becoming more verbose because of the need to support multiple interfaces such as `goyave.PreWriter`, `io.Closer` and now `http.Flusher`. It would be a convenient improvement to provide a `BaseWriter` (or `CommonWriter`) to use with composition and that implements a proper default behavior for all these interfaces. 

[Chained writers docs](https://goyave.dev/basics/response.html#chained-writers).

System-Glitch:
## References

**Issue(s):** closes #215

## Description

- Make `Response` implement `http.Flusher`
- Call `PreWrite` only once on the first `Write`
- Add `CommonWriter` to reduce chained writers boilerplate
- Use `CommonWriter` for log and compress middleware

### DIFF
--- a/middleware/compress/compress.go
+++ b/middleware/compress/compress.go
@@ -27,29 +27,37 @@ type Encoder interface {
 }
 
 type compressWriter struct {
-	io.WriteCloser
-	http.ResponseWriter
-	childWriter io.Writer
+	goyave.CommonWriter
+	responseWriter http.ResponseWriter
+	childWriter    io.Writer
 }
 
 func (w *compressWriter) PreWrite(b []byte) {
 	if pr, ok := w.childWriter.(goyave.PreWriter); ok {
 		pr.PreWrite(b)
 	}
-	h := w.ResponseWriter.Header()
+	h := w.responseWriter.Header()
 	if h.Get("Content-Type") == "" {
 		h.Set("Content-Type", http.DetectContentType(b))
 	}
 	h.Del("Content-Length")
 }
 
-func (w *compressWriter) Write(b []byte) (int, error) {
-	n, err := w.WriteCloser.Write(b)
-	return n, errors.New(err)
+func (w *compressWriter) Flush() error {
+	if err := w.CommonWriter.Flush(); err != nil {
+		return errors.New(err)
+	}
+	switch flusher := w.childWriter.(type) {
+	case goyave.Flusher:
+		return errors.New(flusher.Flush())
+	case http.Flusher:
+		flusher.Flush()
+	}
+	return nil
 }
 
 func (w *compressWriter) Close() error {
-	err := errors.New(w.WriteCloser.Close())
+	err := errors.New(w.CommonWriter.Close())
 
 	if wr, ok := w.childWriter.(io.Closer); ok {
 		return errors.New(wr.Close())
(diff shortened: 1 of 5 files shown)


response.go
```diff
@@ -128,6 +187,25 @@ func (r *Response) Cookie(cookie *http.Cookie) {
...
+// If the response headers have not been written already, `PreWrite()` will
+// be called with an empty byte slice.
+func (r *Response) Flush() {
+	if !r.wroteHeader {
+		r.PreWrite([]byte{})
+	}
+	switch flusher := r.writer.(type) {
+	case Flusher:
+		if err := flusher.Flush(); err != nil {
+			r.server.Logger.Error(errorutil.New(err))
```
ccoVeille, line 202: Does the logger react to the callers stored in the  errorutil.*Error ?

I looked at goyave/slog but found nothing.

If not, it could be simplified

middleware/compress/compress.go
```diff
@@ -106,8 +114,8 @@ func (m *Middleware) Handle(next goyave.Handler) goyave.Handler {
 
 		respWriter := response.Writer()
 		compressWriter := &compressWriter{
```
ccoVeille, line 116: Maybe a NewCompressWriter would help. It could be clearer maybe.

I'm unsure

response.go
System-Glitch: Yes it does.

EdmondFrank:
Awesome, This is exactly what I need right now!

response.go
```diff
@@ -128,6 +187,25 @@ func (r *Response) Cookie(cookie *http.Cookie) {
...
 }
 
+// Flush sends any buffered data to the client if the underlying
+// writer implements `goyave.Flusher`.
+//
+// If the response headers have not been written already, `PreWrite()` will
+// be called with an empty byte slice.
+func (r *Response) Flush() {
+	if !r.wroteHeader {
+		r.PreWrite([]byte{})
```
ccoVeille, line 197: I'm a bit surprised by this line, because it will call prewrite, that will indeed send the header, but I feel strange to PreWrite during flush

Shouldn't you simply call WriteHeader ?
```suggestion
		r.WriteHeader(r.status)
```

response.go
```diff
@@ -81,10 +138,12 @@ func (r *Response) reset(server *Server, request *Request, writer http.ResponseW
...
-		pr.PreWrite(b)
+	if r.empty {
+		if pr, ok := r.writer.(PreWriter); ok {
+			pr.PreWrite(b)
+		}
 	}
+	r.empty = false
 	if !r.wroteHeader {
 		if r.status == 0 {
 			r.status = http.StatusOK
```
ccoVeille, line 149:
```suggestion
	if !r.IsHeaderWritten {
```

ccoVeille: I would move the check if 0, then OK into `r.WriteHeader`

```go
if status == 0 {
   status =  http.StatusOK
}
```

ccoVeille (approved):

response.go
System-Glitch: A flush will trigger a write so we want to trigger `PreWrite` as well. It's only relevant if you call `Flush` without having written anything (no prior call to `Write`).

Issue #195, opened 2024-06-11
Clickhouse database dialect


System-Glitch:
Create a [new database dialect](https://goyave.dev/basics/database.html#drivers) to support Gorm's official [Clickhouse driver](https://github.com/go-gorm/clickhouse).

AidanKenney:
Hey @System-Glitch, I'd be happy to give this a shot.

System-Glitch:
Hello! I'm glad to hear that! If you need assistance, feel free to ask here or on our discord server.

PR #219, opened 2024-07-15
#195 - Add Support for Clickhouse

AidanKenney:
## References

Issue(s):#195  

## Description

Add support for Clickhouse driver using GORM


System-Glitch (approved):

System-Glitch:
Thank you very much for your contribution! Can you check if the `Exists`/`Unique` validator works with clickhouse please?
I don't think it will but it's not a problem for now, another issue will be created if needed.

AidanKenney:
Sure thing @System-Glitch! And I haven't been able to get the clickhouse dialect to work in `unique_test.go` yet, I think the tests will need a bit of refactoring for it to work. Do you want me to try to get that working before moving ahead with this?

System-Glitch:
> Sure thing @System-Glitch! And I haven't been able to get the clickhouse dialect to work in `unique_test.go` yet, I think the tests will need a bit of refactoring for it to work. Do you want me to try to get that working before moving ahead with this?

As you prefer. If you're not willing to do it, this can be done in another issue later on.

AidanKenney:
Let me take another shot at it, I'll follow up if I'm not able to get anywhere

validation/unique_test.go
```diff
@@ -370,6 +370,7 @@ func TestUniqueArrayValidator(t *testing.T) {
 			{dialect: "mysql", expected: "WITH ctx_values(id, i) AS (SELECT * FROM (VALUES ROW(2,0),ROW(7,1),ROW(6,2)) t) SELECT i FROM ctx_values LEFT JOIN `models` ON `models`.`name` = ctx_values.id WHERE `models`.`name` IS NOT NULL"},
 			{dialect: "postgres", expected: "WITH ctx_values(id, i) AS (SELECT * FROM (VALUES (2,0::int),(7,1::int),(6,2::int)) t) SELECT i FROM ctx_values LEFT JOIN `models` ON `models`.`name` = ctx_values.id WHERE `models`.`name` IS NOT NULL"},
 			{dialect: "mssql", expected: "WITH ctx_values(id, i) AS (SELECT * FROM (VALUES (2,0),(7,1),(6,2)) t(id,i)) SELECT i FROM ctx_values LEFT JOIN `models` ON `models`.`name` = ctx_values.id WHERE `models`.`name` IS NOT NULL"},
+			{dialect: "clickhouse", expected: "WITH ctx_values(id, i) AS (SELECT * FROM (VALUES 'a UInt64, b UInt64', (2,0),(7,1),(6,2)) t) SELECT i FROM ctx_values LEFT JOIN `models` ON `models`.`name` = ctx_values.id WHERE `models`.`name` IS NOT NULL"},
```
System-Glitch, line 373: I tested the query on a running Clickhouse docker image and it doesn't work. The following query somewhat works though:
```sql
WITH ctx_values AS
    (
        SELECT *
        FROM VALUES('id UInt64, i UInt64', (2, 0), (7, 1), (6, 2))
    )
SELECT i
FROM ctx_values
LEFT JOIN users ON users.id = ctx_values.id
WHERE users.id IS NOT NULL
```
The query has a valid syntax, however it doesn't work as we want it to work. In this engine, it looks like the type's **zero value** is returned instead of `NULL` when the join doesn't match the `ON` condition. I tested it like so:
```
WITH ctx_values AS
    (
        SELECT *
        FROM VALUES('id UInt64, i UInt64', (1, 0), (2, 1), (4, 2))
    )
SELECT
    id,
    i,
    users.id,
    users.email
FROM ctx_values
LEFT JOIN users ON users.id = ctx_values.id
WHERE users.id IS NOT NULL

Query id: 4ab27032-b547-4b22-9427-6c63f8d0fb61

   ┌─id─┬─i─┬─users.id─┬─email─┐
1. │  1 │ 0 │        1 │ k     │
2. │  2 │ 1 │        0 │       │
3. │  4 │ 2 │        0 │       │
   └────┴───┴──────────┴───────┘

3 rows in set. Elapsed: 0.002 sec. 
```
This query should only return the first row but it doesn't because users.id defaults to 0 instead of NULL.

---

It's getting very different from the rest of the database engine so we probably should create a separate function to avoid overcomplicating `buildQuery()`:

```go
func (v *ExistsArrayValidator[T]) buildQuery(values []T, condition bool) *gorm.DB {
	dbType := v.Config().GetString("database.connection")
	if dbType == "clickhouse" {
		return v.buildClickhouseQuery(values, condition)
	}
        //...
```

Another important thing: the `id` column in the `ctx_values`  won't always be `uint64`. It will be of type `T`, the generic parameter of `ExistsArrayValidator`. I'm not sure how that works in ClickHouse. Maybe there's a way to get the clickhouse type name from a Go type in the [official lib](https://github.com/ClickHouse/clickhouse-go)? 

This really isn't very practical...

System-Glitch (changes requested):

validation/unique_test.go
AidanKenney: Thanks for this investigation, I see what you mean and had been wondering about this myself. 

Regarding the UInt64 type defaulting to null, it looks like we would get the desired behavior from an `INNER JOIN` rather than a `LEFT JOIN`: 

`WITH ctx_values AS
    (
        SELECT *
        FROM VALUES('id UInt64, i UInt64', (1, 0), (2, 1), (4, 2))
    )
SELECT
    id,
    i,
    test.users.id,
    test.users.email
FROM ctx_values
INNER JOIN test.users ON test.users.id = ctx_values.id
WHERE test.users.id IS NOT NULL

Query id: 2e169b8f-6fcf-4e7a-9fcd-689e5ff42031

   ┌─id─┬─i─┬─users.id─┬─email─┐
1. │  1 │ 0 │        1 │ k     │
   └────┴───┴──────────┴───────┘

1 row in set. Elapsed: 0.010 sec.`

This could also be accomplished by a `LEFT JOIN` when the JOIN setting "[join_use_nulls](https://clickhouse.com/docs/en/operations/settings/settings#join_use_nulls)" is set to 1: 

`SET join_use_nulls = 1

Query id: 3a88865e-25fe-4ade-aeea-87f24c956eee

Ok.

0 rows in set. Elapsed: 0.004 sec.

...

WITH ctx_values AS
    (
        SELECT *
        FROM VALUES('id UInt64, i UInt64', (1, 0), (2, 1), (4, 2))
    )
SELECT
    id,
    i,
    test.users.id,
    test.users.email
FROM ctx_values
LEFT JOIN test.users ON test.users.id = ctx_values.id
WHERE test.users.id IS NOT NULL

Query id: d55eadea-2bcb-455b-873a-77205f4d0bfc

   ┌─id─┬─i─┬─users.id─┬─email─┐
1. │  1 │ 0 │        1 │ k     │
   └────┴───┴──────────┴───────┘

1 row in set. Elapsed: 0.009 sec.`

Do you have any opinions on how the best way to deal with this? 

I will keep looking into the existence of a reliable Go type <> Clickhouse type mapping.

AidanKenney: Ah, looks like this might be what we needed: https://github.com/ClickHouse/clickhouse-go/blob/main/TYPES.md

validation/unique.go
```diff
@@ -144,6 +148,68 @@ func (v *ExistsArrayValidator[T]) buildQuery(values []T, condition bool) *gorm.D
...
+	reflect.TypeOf(int64(0)):   "Int64",
+	reflect.TypeOf(int32(0)):   "Int32",
+	reflect.TypeOf(int16(0)):   "Int16",
+	reflect.TypeOf(int8(0)):    "Int8",
+	reflect.TypeOf(int(0)):     "Int64",
+	reflect.TypeOf(float32(0)): "Float32",
+	reflect.TypeOf(float64(0)): "Float64",
+	reflect.TypeOf(""):         "String",
+	reflect.TypeOf(true):       "Bool",
+}
```
System-Glitch, line 211: Nice job, that will do. This way we don't have to import an additional package.
I am just worried that it won't work with types like uuid, time, and the other type-conversion validators. Let's add the most common ones and let users handling the custom ones with the transform function.

validation/unique.go
```diff
@@ -144,6 +148,68 @@ func (v *ExistsArrayValidator[T]) buildQuery(values []T, condition bool) *gorm.D
...
+		questionMarks = append(questionMarks, "?")
+		var transformedValue any = val
+		if v.Transform != nil {
+			transformedValue = v.Transform(val)
+		}
+		params = append(params, gorm.Expr(
+			"(?,?)",
+			transformedValue, i))
+		if paramZeroType == "" {
+			paramZeroType = clickhouseTypes[reflect.TypeOf(transformedValue)]
```
System-Glitch, line 169: This can be done once outside the loop. It would be cleaner in my opinion:
```go
var zeroVal T
paramType, ok := clickhouseTypes[reflect.TypeOf(zeroVal)
```
You should check if the value is supported, otherwise [add an error](https://pkg.go.dev/goyave.dev/goyave/v5@v5.1.1/validation#Context.AddError) to the validation context if no `Transform` function is provided.

`transformedValue` is a `gorm.Expr` so your reflect call here will never work if you provide a transform function.

AidanKenney: So if the value is not supported within the `clickhouseTypes` map, but there is a Transform function provided, will we be able to determine what `paramType` should be?

validation/unique.go
```diff
@@ -144,6 +148,68 @@ func (v *ExistsArrayValidator[T]) buildQuery(values []T, condition bool) *gorm.D
...
+			paramOneType = clickhouseTypes[reflect.TypeOf(i)]
+		}
+	}
+
+	db := v.DB()
+	table := db.Statement.Quote(v.Table)
+	column := db.Statement.Quote(v.Column)
+
+	sql := fmt.Sprintf(
+		"WITH ctx_values(id, i) AS (SELECT * FROM (VALUES 'id %s, i %s', %s)) SELECT i FROM ctx_values INNER JOIN %s ON %s.%s = ctx_values.id WHERE %s.%s IS %s NULL",
```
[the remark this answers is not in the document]
AidanKenney: Yes this will result in `SELECT * FROM (VALUES 'id Int64, i Int64', (2,0),(7,1),(6,2))` which is valid Clickhouse syntax.

validation/unique.go
System-Glitch: Transform is an option for special cases. For example for Postgres it allows to do the following:
```go
UniqueArray("table", "column", func(val int64) clause.Expr {
	return gorm.Expr("?::bigint", val)
})
```
```go
UniqueArray("table", "column", func(val string) clause.Expr {
	return gorm.Expr("LOWER(?)", val)
})
```

So in the case of clickhouse, this would allow developers to cast raw values into special types that we don't support directly.

System-Glitch (changes requested):

System-Glitch (approved):

System-Glitch:
Merged. Thank you very much! 🎉

Issue #221, opened 2024-08-07
Not found error skipped in JWT Controller


akhedrane:
Hello, 
Not found error skipped in JWT Controller that caused reflection error on line 123
https://github.com/go-goyave/goyave/blob/3e4f090a2c5ac7c02cb17cf47a64d6958f9f5ffa/auth/jwt_controller.go#L123

System-Glitch:
Hello. Thank you for your report. Can you include a stack trace please?

akhedrane:
the stack trace:


> reflect: call of reflect.Value.FieldByName on zero Value
trace: 
reflect.flag.mustBe
        /Go/src/reflect/value.go:233
reflect.Value.FieldByName
        /Go/src/reflect/value.go:1361
goyave.dev/goyave/v5/auth.(*JWTController[...]).Login
        /go/pkg/mod/goyave.dev/goyave/v5@v5.2.0/auth/jwt_controller.go:123
goyave.dev/goyave/v5.(*validateRequestMiddleware).Handle.func1
       /go/pkg/mod/goyave.dev/goyave/v5@v5.2.0/middleware.go:202
goyave.dev/goyave/v5/middleware/parse.(*Middleware).Handle.func1
        /go/pkg/mod/goyave.dev/goyave/v5@v5.2.0/middleware/parse/parse.go:55
goyave.dev/goyave/v5/middleware/parse.(*Middleware).Handle.func1
        /go/pkg/mod/goyave.dev/goyave/v5@v5.2.0/middleware/parse/parse.go:97
goyave.dev/goyave/v5/auth.(*Handler[...]).Handle.func1
        /go/pkg/mod/goyave.dev/goyave/v5@v5.2.0/auth/authenticator.go:70
goyave.dev/goyave/v5/log.(*AccessMiddleware).Handle.func1
        /go/pkg/mod/goyave.dev/goyave/v5@v5.2.0/log/log.go:100
goyave.dev/goyave/v5.(*corsMiddleware).Handle.func1
        /go/pkg/mod/goyave.dev/goyave/v5@v5.2.0/middleware.go:237
goyave.dev/goyave/v5.(*languageMiddleware).Handle.func1
        /go/pkg/mod/goyave.dev/goyave/v5@v5.2.0/middleware.go:118
goyave.dev/goyave/v5.(*recoveryMiddleware).Handle.func1
        /go/pkg/mod/goyave.dev/goyave/v5@v5.2.0/middleware.go:90
[...]

PR #224, opened 2024-08-12
#221 Auth: Fix JWT controller gorm.ErrRecordNotFound handling

System-Glitch:
## References

**Issue(s):** #221 

## Description

Fixed a reflection error in the Login handler of the JWT Controller. If the user service returns `gorm.ErrRecordNotFound`, a `401 Unauthorized` response will be returned.

### DIFF
--- a/auth/jwt_controller_test.go
+++ b/auth/jwt_controller_test.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"gorm.io/gorm"
 	"goyave.dev/goyave/v5"
 	"goyave.dev/goyave/v5/slog"
 	"goyave.dev/goyave/v5/util/testutil"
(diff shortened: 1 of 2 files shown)


auth/jwt_controller_test.go
```diff
@@ -95,6 +96,32 @@ func TestJWTController(t *testing.T) {
...
+		}
+		body, err := json.Marshal(data)
+		require.NoError(t, err)
+		request := httptest.NewRequest(http.MethodPost, "/login", bytes.NewReader(body))
+		request.Header.Set("Content-Type", "application/json")
+		resp := server.TestRequest(request)
+		assert.Equal(t, http.StatusUnauthorized, resp.StatusCode)
+		respBody, err := testutil.ReadJSONBody[map[string]string](resp.Body)
+		assert.NoError(t, resp.Body.Close())
+		require.NoError(t, err)
```
ccoVeille, line 121: I would invert these two lines

```suggestion
		require.NoError(t, err)
		assert.NoError(t, resp.Body.Close())
```

Because if an error occurs, resp will be nil, so resp.Body.Close() would panic

System-Glitch: We are testing the error from `testutil.ReadJSONBody` here. The `resp` won't be `nil` at this point.

Issue #228, opened 2024-09-19
Testing: easy way to discard logs


System-Glitch:
## Proposal

Add a `slog.Discard` logger that outputs to `io.Discord`. This will make test servers easier to generate:
```go
server := testutil.NewTestServerWithOptions(t, goyave.Options{Logger: slog.Discard)})
```
Maybe use the discard logger by default in test servers.

PR #229, opened 2024-09-20
#228 - Feat: create slog discard logger

SallesCosta:
## References

**Issue(s):** #228    
**Discussion:**  no discussion

## Description

Created the logger with `io.Discard` and set it as the default in NewTestServerWithOptions if no logger is provided as a parameter.

### Possible drawbacks

I noticed that `io.Discard` is already used [here in log_bench_test.go](https://github.com/go-goyave/goyave/blob/2ce8e01d18639caf00098200b28d973b2420d7b4/log/log_bench_test.go#L17) in a `slog.NewHandler`. In the implementation of `slog.Discard` I chose to use `slog.NewTextHandler` because I understand that it achieves the same objective, however it is more evident that it is a text.

### DIFF
--- a/slog/slog.go
+++ b/slog/slog.go
@@ -3,6 +3,7 @@ package slog
 import (
 	"context"
 	"fmt"
+	"io"
 	"reflect"
 	"runtime"
 	"time"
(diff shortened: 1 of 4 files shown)


System-Glitch (approved):

System-Glitch:
Merged! Thanks a lot! 🎉 
I will update the docs (on goyave.dev).

Issue #205, opened 2024-06-11
walk.Path escape character


System-Glitch:
## Proposal

It is currently impossible to use `*` or `[]` in paths parsed with the `util/walk` package. We could add a way to escape these characters in the parser.

achimoraites:
Hello @System-Glitch !

While I am new to Go, I am interested in this project

System-Glitch:
Hi @achimoraites and thank you for your interest. Feel free to ask me if you need any help or guidance.

PR #236, opened 2025-01-03
fix(util):  add escape system so '*', '[]' can be escaped

achimoraites:
## References

**Issue(s):** #205

## Description

Add escape system to enable `walk.Path` to use `*`  `[` `]` in paths 

Examples:
- `object\*field`
- `array[][].field\\[]`
- `array[][].field\\[tx\\]`

### DIFF
--- a/util/walk/walk.go
+++ b/util/walk/walk.go
@@ -357,16 +357,18 @@ func (p *Path) setAllMissingIndexes() {
 //
 // Example paths:
 //
-//	name
-//	object.field
-//	object.subobject.field
-//	object.*
-//	object.array[]
-//	object.arrayOfObjects[].field
-//	[]
-//	[].field
+//			name
+//			object.field
+//			object.subobject.field
+//			object.*
+//			object.array[]
+//			object.arrayOfObjects[].field
+//			[]
+//			[].field
+//		 	object*field
+//	     	object.field\[]
+//	     	object.field\[text\]
 func Parse(p string) (*Path, error) {
-	// TODO add escape system so '*', '[]' can be escaped
 	rootPath := &Path{}
 	path := rootPath
 
@@ -404,6 +406,7 @@ func Parse(p string) (*Path, error) {
 				path = path.Next
 			}
 		default:
+			t = removeEscapeChars(t)
 			path.Name = &t
 		}
 	}
@@ -436,6 +439,12 @@ func Depth(p string) uint {
 }
 
 func createPathScanner(path string) *bufio.Scanner {
+	escape := map[rune]bool{
+		'*': true,
+		'[': true,
+		']': true,
+	}
+
 	scanner := bufio.NewScanner(strings.NewReader(path))
 	split := func(data []byte, atEOF bool) (int, []byte, error) {
 		if len(path) == 0 || path[0] == '.' {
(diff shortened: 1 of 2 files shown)


util/walk/walk.go
```diff
@@ -436,6 +439,12 @@ func Depth(p string) uint {
 }
 
 func createPathScanner(path string) *bufio.Scanner {
+	escape := map[rune]bool{
```
System-Glitch, line 442: For a tiny memory optimization, can you change this to a `map[rune]struct{}` please?

util/walk/walk.go
```diff
@@ -447,6 +456,13 @@ func createPathScanner(path string) *bufio.Scanner {
...
 			if i+width < len(data) {
 				next, _ := utf8.DecodeRune(data[i+width:])
+				// case: escape chars
+				_, escapeNext := escape[next]
+				if r == '\\' && escapeNext {
+					// Skips the next character
+					i += width
+					continue
+				}
 				if isValidSyntax(r, next) {
```
System-Glitch, line 466: You need to check in this function that a backslash (`\`) is followed by either another backslash or the characters that can be escaped. In other words, you should also escape the backslash character. `path\to\element` should be an invalid path, while `path\\to\\element` should be valid.

This will create a behavior that is more intuitive and closer to what developers are used to when escaping characters in general.

This function is badly named, it should be `isSyntaxInvalid` or something like that.

System-Glitch (changes requested):
Thank you very much for your PR! 

The approach is good, but there are some cases that are not covered in your current implementation.

`path.String()` should also return the path with the escape characters (`\`) shown to avoid confusion.

I'm not sure exactly how we should do that. Maybe we can add a `Original string` field in the `Path` struct. This would break a ton of tests though.

Alternatively, we could keep the original path segment in the `Name` field and `removeEscapeChars` when walking in `path.Walk`. Other uses of this field (in the validation package) will need to be updated.

If you have a better suggestion, feel free to suggest!

achimoraites:
> Thank you very much for your PR!
> 
> The approach is good, but there are some cases that are not covered in your current implementation.
> 
> `path.String()` should also return the path with the escape characters (`\`) shown to avoid confusion.
> 
> I'm not sure exactly how we should do that. Maybe we can add a `Original string` field in the `Path` struct. This would break a ton of tests though.
> 
> Alternatively, we could keep the original path segment in the `Name` field and `removeEscapeChars` when walking in `path.Walk`. Other uses of this field (in the validation package) will need to be updated.
> 
> If you have a better suggestion, feel free to suggest!

@System-Glitch
:thinking: I will will let you know when I have have a closer look at the affected code.

Thanks for your feedback!

System-Glitch:
@achimoraites Do you mind if I continue working on your branch for the last minor requirements?

System-Glitch:
There are a some issues and edge cases. I'm working on it:
- The scanner stops splitting the tokens after the first escaped character.
- `removeEscapeChars` is slow because it rebuilds a new string replacer in every call.
- The error messages when the path contains escaped characters doesn't match the actual string (the escaped characters appear escaped too so they are double-escaped)
- Walking a path containing an escaped wildcard results in a stack overflow error. I will add a way to be able to differentiate between a wildcard and an asterisk.
- The `Depth()` function doesn't take escaped characters into account. This could in turn break composition in validation.
- Minor issues when translating a field name containing dots in validation error messages.
- `RuleSet` parsing didn't take into account escaped characters, which resulted in weird behavior when working with arrays.
- Single value arrays conversion was broken in validation because it didn't take into account escaped characters.

That may look like a lot but most of them are quite easy to fix actually.

System-Glitch:
@achimoraites I just pushed the additional changes. I think it's ready now. Feel free to have a quick look if you want, I'll merge this in a few days I think.

System-Glitch:
Merged! 🎉 Thanks a lot for your contribution. I'll need a bit more time to update documentation before releasing it though.

PR #259, opened 2025-04-03
validation: fix incorrect validation order w/ composition for array elements + panic on duplicate path

### LINKED ISSUE #248: Validation: incorrect validation order when using composition for an array element
## Bug report

**Version:** v5.5.6
**Environment:** `go version go1.24.0 linux/amd64`
**Reproduces:** always

### Reproduce

```go
func RootValidation(r *goyave.Request) v.RuleSet {
	return v.RuleSet{
		{Path: v.CurrentElement, Rules: v.List{v.Required(), v.Object()}},
		{Path: "field1", Rules: v.List{v.Required(), v.String()}},
		{Path: "array", Rules: v.List{v.Required(), v.Array()}},
		{Path: "array[]", Rules: ElementValidation(r)},
	}
}

func ElementValidation(_ *goyave.Request) v.RuleSet {
	return v.RuleSet{
		{Path: v.CurrentElement, Rules: v.List{v.Object()}},
		{Path: "element_field", Rules: v.List{v.Required(), v.String()}},
	}
}
```

The order of validation will be:
1. `array[].element_field`
2. Root object
3. `field1`
4. `array[]` (arrays are validated recursively, so this is expected that `array[]` is before `array`)
5. `array`

### Expected result

The order of validation **should** be:
1. Root object
2. `field1`
3. `array[]`
4. `array`
5. `array[].element_field`

### LINKED ISSUE #249: Validation: cannot validate the same path twice for array elements
## Bug report

**Version:** v5.5.6
**Environment:** `go version go1.24.0 linux/amd64`
**Reproduces:** always

### Reproduce

```go
func RootValidation(r *goyave.Request) v.RuleSet {
	return v.RuleSet{
		{Path: v.CurrentElement, Rules: v.List{v.Required(), v.Object()}},
		{Path: "field1", Rules: v.List{v.Required(), v.String()}},
		{Path: "array", Rules: v.List{v.Required(), v.Array()}},
		{Path: "array[]", Rules: ElementValidation(r)},
		{Path: "array2", Rules: v.List{v.Required(), v.Array()}},
		{Path: "array2[]", Rules: v.List{v.Int32()}},
		{Path: "array2[]", Rules: v.List{v.Int64()}},
		{Path: v.CurrentElement, Rules: v.List{RootPostValidation()}},
	}
}

func ElementValidation(_ *goyave.Request) v.RuleSet {
	return v.RuleSet{
		{Path: v.CurrentElement, Rules: v.List{v.Object()}},
		{Path: "element_field", Rules: v.List{v.Required(), v.String()}},
		{Path: v.CurrentElement, Rules: v.List{ElementPostValidation()}},
	}
}
```

In `RootValidation` we want two passes on the root object. The use case is when you want to validate something related to multiple fields in the object after those fields were validated. Because the order of validation is guaranteed, being able to pass multiple times on the same path can be very useful.

Here, it works as expected for the root object in `RootValidation`. However:
- in the case of `array2[]`, only `Int64` is kept, while `Int32` is dropped.
- in the case of composition, the `Object` validator is dropped for the "element" in `ElementValidation`.

### Expected result

It should be possible to validate an array element several times, just like other fields.

System-Glitch:
## References

**Issue(s):** 
- fixes #248
- fixes #249 

## Description

- Fixed incorrect validation order when using composition for array elements. See #248 for more details. This was achieved by using a map storing the array `*Field` instead of using lookups and removing the sorting step. 
- Now panic when a path is repeated in a rule set. See [this comment](https://github.com/go-goyave/goyave/issues/249#issuecomment-2775827035)

### Possible drawbacks

- More allocations per RuleSet. Memory usage slightly increased. Execution speed is very slightly slower but expected to be faster for large rule sets because the complexity of `AsRules()` has been reduced. Overall, the impact is acceptable.
- Existing code in projects using previous versions of Goyave may panic now. I don't consider it a breaking change since this behavior has always been buggy and never fully assumed to be a valid use-case.

### DIFF
--- a/validation/ruleset.go
+++ b/validation/ruleset.go
@@ -6,6 +6,7 @@ import (
 	"slices"
 
 	"github.com/samber/lo"
+	"goyave.dev/goyave/v5/util/errors"
 	"goyave.dev/goyave/v5/util/walk"
 )
 
@@ -103,7 +104,6 @@ func (l List) convert(path string, field *FieldRules, prefixDepth uint) Rules {
 // FieldRules structure associating a path (see `walk.Path`) identifying a field
 // with a `FieldRulesApplier` (a `List` of rules or another `RuleSet` via composition).
 type FieldRules struct {
-	// TODO what behavior if there are duplicates? If it ever becomes a problem, can probably merge the Lists. But it's unnecessary for now.
 	Rules FieldRulesConverter
 	Path  string
 }
@@ -152,6 +152,8 @@ func (r RuleSet) asRulesWithPrefix(prefix string) Rules {
 		}
 	}
 
+	rules.checkDuplicates()
+
 	for {
 		arrayElement, index, ok := lo.FindIndexOf(rules, func(f *Field) bool {
 			p := f.Path
@@ -174,7 +176,7 @@ func (r RuleSet) asRulesWithPrefix(prefix string) Rules {
 		parentArrayElement, parentFound := arrays[parentArrayPathStr]
 
 		rules = slices.Delete(rules, index, index+1)
-		if parentFound { // Should never be false because we injected array parents.
+		if parentFound { // Should never be false because we injected array parents and there are no duplicates.
 			arrayElement.Path = &walk.Path{Type: walk.PathTypeArray, Next: &walk.Path{}}
 			parentArrayElement.Elements = arrayElement
 		}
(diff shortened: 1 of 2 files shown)


System-Glitch:
Keeping this PR as a draft for now, as I will probably include the fix for #249 in it as well.

Issue #279, opened 2025-09-30
Ability to provide a net.ListenConfig


System-Glitch:
## Proposal

Currently, the net listener is opened with the default `net.ListenConfig` and there is no way to customize that. It can be useful if the keep-alives need to be tweaked for optimal performance in large traffic scenarios.

In `goyave.Options`, we should add an optional `net.ListenConfig` field that would be used in `server.Start()` if provided.

PR #280, opened 2025-09-30
Add custom ListenConfig support for server

Supakornn:
## References

**Issue(s):** #279   
**Discussion:** #yyy  

## Description

Adds support for customizing the net.ListenConfig when starting the server. 

Usage Example
```go
customListenConfig := &net.ListenConfig{
    KeepAlive: 30 * time.Second,
}

server, err := goyave.New(goyave.Options{
    ListenConfig: customListenConfig,
})
```

### DIFF
--- a/server_test.go
+++ b/server_test.go
@@ -694,6 +694,58 @@ func TestServer(t *testing.T) {
 			assert.Equal(t, "cannot start the server, context is canceled", err.Error())
 		}
 	})
+
+	t.Run("StartWithCustomListenConfig", func(t *testing.T) {
+		cfg := config.LoadDefault()
+		cfg.Set("server.port", 0)
+
+		// Create a custom ListenConfig with custom keep-alive settings
+		customListenConfig := &net.ListenConfig{
+			KeepAlive: 1 * time.Minute, // Custom keep-alive duration
+		}
+
+		server, err := New(Options{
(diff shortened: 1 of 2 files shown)


server_test.go
```diff
@@ -694,6 +694,58 @@ func TestServer(t *testing.T) {
 			assert.Equal(t, "cannot start the server, context is canceled", err.Error())
 		}
 	})
+
+	t.Run("StartWithCustomListenConfig", func(t *testing.T) {
```
System-Glitch, line 698: That's too bad there is no way for us to check the `*net.TCPListener.lc` in this test. Currently, we know the server works but the test doesn't ensure the listen config is actually used. We could use reflection and `unsafe.Pointer` but I'd rather not.

Can you add a second test with a `ListenConfig.Control` function that always returns an error, and check for this error in the assertions? This will ensure the given ListenConfig is actually used.

Can you please add an assertion on `server.listenConfig` in the `"New"` test to check the value from the config is copied to the server? (The same way I do with the HTTP2 config)
```go
assert.Same(t, customListenConfig, s.server.listenConfig)
```

System-Glitch (changes requested):
Thanks a lot for your contribution! Apart from the minor missing test cases, LGTM 👍🏻

PR #286, opened 2026-03-02
fix: the MIME type of an SVG file should be image/svg+xml

fnoopv:
## References

**Issue(s):** #xxx   
**Discussion:** #yyy  

## Description
The MIME type of an SVG file should be image/svg+xml, otherwise, the front-end cannot render the image based on the MIME type.

### DIFF
--- a/util/fsutil/fsutil.go
+++ b/util/fsutil/fsutil.go
@@ -149,10 +149,26 @@ func DetectContentType(r io.Reader, fileName string) (string, error) {
 }
 
 func hasSVGSignature(buffer []byte) bool {
-	content := strings.TrimSpace(strings.ToLower(string(buffer)))
-	if after, ok := strings.CutPrefix(content, "\ufeff"); ok {
-		content = strings.TrimSpace(after)
+	start := -1
+	for i, b := range buffer {
+		if b == '<' {
+			start = i
+			break
+		}
 	}
+	if start == -1 {
+		return false
+	}
+
+	raw := buffer[start:]
+	normalized := make([]byte, 0, len(raw))
+	for _, b := range raw {
+		if b != 0 {
+			normalized = append(normalized, b)
+		}
+	}
+
+	content := strings.TrimSpace(strings.ToLower(string(normalized)))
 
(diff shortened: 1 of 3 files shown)


util/fsutil/fsutil.go
```diff
@@ -148,7 +149,7 @@ func detectContentTypeByExtension(fileName, contentType string) string {
 	for ext, t := range contentTypeByExtension {
 		if strings.HasSuffix(fileName, ext) {
 			tmp := t
-			if i := strings.Index(contentType, ";"); i != -1 {
+			if i := strings.Index(contentType, ";"); i != -1 && t != "image/svg+xml" {
```
System-Glitch: Why do we need this condition?

util/fsutil/fsutil.go
```diff
@@ -138,7 +139,7 @@ func DetectContentType(r io.Reader, fileName string) (string, error) {
 	}
 
 	contentType := http.DetectContentType(buffer)
-	if strings.HasPrefix(contentType, "application/octet-stream") || strings.HasPrefix(contentType, "text/plain") {
+	if strings.HasPrefix(contentType, "application/octet-stream") || strings.HasPrefix(contentType, "text/plain") || strings.HasPrefix(contentType, "text/xml") || strings.HasPrefix(contentType, "application/xml") {
```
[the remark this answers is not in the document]
fnoopv: I will expand sniffing of SVG.

util/fsutil/fsutil.go
```diff
@@ -140,15 +142,46 @@ func DetectContentType(r io.Reader, fileName string) (string, error) {
...
+			content = strings.TrimSpace(content[end+3:])
+			continue
+		}
+		break
+	}
+
+	return strings.HasPrefix(content, "<svg")
+}
+
 func detectContentTypeByExtension(fileName, contentType string) string {
```
System-Glitch: Please add an early return here if `fileName` is empty to avoid extra processing.

System-Glitch (changes requested):

PR #287, opened 2026-06-02
Fix duplicate path join in OS filesystem utility

fnoopv:
## References

**Issue(s):** N/A  
**Discussion:** N/A

## Description

Fix `osfs.FS.FileExists` and `osfs.FS.IsDirectory` when a base directory is configured with `New(...)`.

Previously, both methods called `path.Join(f.dir, name)` before delegating to `f.Stat(name)`. Since `Stat` already resolves the given path relative to the filesystem base directory, the base path was effectively joined twice. This caused lookups such as `New("resources").FileExists("test_file.txt")` and `New("resources").IsDirectory("lang")` to fail unexpectedly.

This PR removes the duplicated `path.Join` in both methods and adds regression tests covering `FileExists` and `IsDirectory` with a configured base directory.

### Usage example

```go
fs := osfs.New("resources")

exists := fs.FileExists("test_file.txt") // true
isDir := fs.IsDirectory("lang")          // true
```

### Possible drawbacks

This change slightly alters the behavior of `FileExists` and `IsDirectory` for callers relying on the previous incorrect path resolution. However, the new behavior is consistent with the rest of the filesystem abstraction and with how `Stat` already handles base directories.

### DIFF
--- a/util/fsutil/osfs/osfs.go
+++ b/util/fsutil/osfs/osfs.go
@@ -72,7 +72,7 @@ func (*FS) Getwd() (string, error) {
 // FileExists returns true if the file at the given path exists and is readable.
 // Returns false if the given file is a directory.
 func (f *FS) FileExists(name string) bool {
-	if stats, err := f.Stat(path.Join(f.dir, name)); err == nil {
+	if stats, err := f.Stat(name); err == nil {
 		return !stats.IsDir()
 	}
 	return false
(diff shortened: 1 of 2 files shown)


System-Glitch:
> This change slightly alters the behavior of FileExists and IsDirectory for callers relying on the previous incorrect path resolution. However, the new behavior is consistent with the rest of the filesystem abstraction and with how Stat already handles base directories.

Any patch affecting code relying on buggy behavior should not be considered a breaking change. This is fine.

System-Glitch (approved):

System-Glitch:
Merged. Thanks!

PR #293, opened 2026-09-01
fix: don't double-panic in recovery middleware on empty error slices

fnoopv:
errors.NewSkip returns nil for nil or empty-slice panic values, so the unconditional type assertion panicked again inside the deferred recovery, breaking the connection instead of returning a 500. Use a comma-ok assertion and add a regression test for panic([]error{}).

## References

no

## Problem
`recoveryMiddleware` did an unconditional type assertion:

    e := errors.NewSkip(err, 4).(*errors.Error)

`errors.NewSkip` returns `nil` for a `nil` reason or an empty error
slice (`panic([]error{})`). In that case the assertion panics again
inside the deferred `recover`, so the connection is abruptly closed by
the `net/http` server instead of returning a `500`.

## Changes
Use a comma-ok type assertion so an unrepresentable panic value
(nil/empty slice) is handled gracefully: the status is still forced to
`500`, but no second panic occurs.

## Tests
Add `empty_slice_panic` regression test asserting the handler no longer
panics, returns `500`, and leaves no error/log behind.

### Possible drawbacks

no

### DIFF
--- a/middleware.go
+++ b/middleware.go
@@ -81,9 +81,11 @@ func (m *recoveryMiddleware) Handle(next Handler) Handler {
 		panicked := true
 		defer func() {
 			if err := recover(); err != nil || panicked {
-				e := errors.NewSkip(err, 4).(*errors.Error) // Skipped: runtime.Callers, NewSkip, this func, runtime.panic
+				e := errors.NewSkip(err, 4) // Skipped: runtime.Callers, NewSkip, this func, runtime.panic
+				if e != nil {
+					response.err = e.(*errors.Error)
+				}
 				m.Logger().Error(e)
-				response.err = e
 				if !response.wroteHeader {
 					response.status = http.StatusInternalServerError // Force status override if the header hasn't been written yet.
 				}
--- a/middleware_test.go
+++ b/middleware_test.go
@@ -172,6 +172,34 @@ func TestRecoveryMiddleware(t *testing.T) {
 		assert.Equal(t, http.StatusInternalServerError, response.status)
 	})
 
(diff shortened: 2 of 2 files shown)


middleware.go
```diff
@@ -81,9 +81,10 @@ func (m *recoveryMiddleware) Handle(next Handler) Handler {
 		panicked := true
 		defer func() {
 			if err := recover(); err != nil || panicked {
-				e := errors.NewSkip(err, 4).(*errors.Error) // Skipped: runtime.Callers, NewSkip, this func, runtime.panic
-				m.Logger().Error(e)
-				response.err = e
+				if e, ok := errors.NewSkip(err, 4).(*errors.Error); ok { // Skipped: runtime.Callers, NewSkip, this func, runtime.panic
+					m.Logger().Error(e)
```
System-Glitch: Even if the error is `nil`, it should be printed in the logs.

middleware_test.go
```diff
@@ -172,6 +172,31 @@ func TestRecoveryMiddleware(t *testing.T) {
...
+
+		request := NewRequest(httptest.NewRequest(http.MethodGet, "/test", nil))
+		response := NewResponse(server, request, httptest.NewRecorder())
+
+		require.NotPanics(t, func() {
+			handler(response, request)
+		})
+
+		assert.Nil(t, response.GetError())
+		assert.Empty(t, logBuffer.String())
```
System-Glitch: This test needs to be updated since we want to print the `nil` error.

System-Glitch (changes requested):
Thanks a lot for finding this edge case!

System-Glitch:
Thanks!

PR #295, opened 2026-09-07
httputil: fix quality value parsing in ParseMultiValuesHeader

MaxFreedomPollard:
## References

**Issue(s):** N/A  
**Discussion:** N/A

## Description

`ParseMultiValuesHeader` matched everything located after the first `;` in a header value against `^q=([01]\.[0-9]{1,3})$` (`util/httputil/httputil.go:17`, used at line 53). When that regex did not match, the priority fell back to `0`, so the value sorted last instead of first. Three shapes that RFC 9110 allows do not match it:

- optional whitespace is allowed after the `;`, e.g. `gzip; q=0.9`
- a quality value may have no fraction, e.g. `gzip;q=1` and `gzip;q=0`
- the `q` parameter is not necessarily the first parameter, e.g. `text/html;charset=utf-8;q=0.9`

Both callers of the function ended up with the client's preference backwards. `Accept-Encoding: gzip;q=1, br;q=0.9` made `compress.Middleware.getEncoder` return `br`, and `Accept-Language: en-US;q=0.5, fr-FR; q=0.9` made `lang.Languages.DetectLanguage` return `en-US`.

The fix walks the parameter list and reads the `q` parameter out of it instead of matching the list as a whole, comparing the parameter name case insensitively as RFC 9110 requires, and the quality value itself is checked without a regex: `strconv.ParseFloat` parses it, then anything outside `0` to `1`, anything that is not made of digits and a single decimal point, and anything with more than three digits after that point is rejected, so the integer forms `q=1` and `q=0` are accepted. A value with no `q` parameter gets a priority of 1, which is the RFC 9110 default and already what a value with no parameters at all got.

The new cases in `util/httputil/httputil_test.go`, the two new `TestEncoderPriority` cases and the new `DetectLanguage` assertion all fail on `master` and pass with the change. `go test -race -coverpkg=./... ./...` passes on the whole module and `golangci-lint run` (v2.13.1, repo config) reports 0 issues. `BenchmarkParseMultiValuesHeader` goes from 4 allocs/op to 2 allocs/op, since scanning the parameters no longer builds a submatch slice.

### Possible drawbacks

A value that carries parameters but no `q`, such as `text/html;charset=utf-8`, now has a priority of 1 instead of 0, so it can outrank values it used to sort behind. That is the RFC 9110 default and it matches how a value with no parameters is already handled, but it is a behaviour change for anyone parsing an `Accept` header that has media type parameters.

### DIFF
--- a/util/httputil/httputil.go
+++ b/util/httputil/httputil.go
@@ -1,7 +1,6 @@
 package httputil
 
 import (
-	"regexp"
 	"sort"
 	"strconv"
 	"strings"
@@ -14,17 +13,35 @@ type HeaderValue struct {
 	Priority float64
 }
 
-var qualityValueRegex = regexp.MustCompile(`^q=([01]\.[0-9]{1,3})$`)
-
 // ParseMultiValuesHeader parses multi-values HTTP headers, taking the
 // quality values into account. The result is a slice of values sorted
 // according to the order of priority.
+// If the input is empty, returns an empty slice.
 //
-// The input is trimmed. If the input is empty, returns an empty slice.
+// The priority of a value is the value of its "q" parameter, the quality
+// value defined by RFC 9110 section 12.4.2: a number between 0 and 1 with
+// at most three digits after the decimal point. Values without a "q"
+// parameter have a priority of 1, the default given by the RFC. If the
+// quality value cannot be parsed, the priority is 0.
 //
-// See: https://developer.mozilla.org/en-US/docs/Glossary/Quality_values
+// Parameters other than "q" are supported, for example the "charset" of a
+// media type. They are ignored: they don't change the priority, and they are
+// not part of the returned value, which only contains what is located before
+// the first ";".
 //
-// For the following header:
+// The input and each segment are trimmed. This includes the header value, the
+// parameter name and the value of the "q" parameter.
+// "text/html ; q = 0.5" is therefore parsed the same way
+// as "text/html;q=0.5".
+//
+// The "q" parameter name is case-insensitive. "Q=0.5" is accepted.
+//
+// See:
+//
+//   - https://datatracker.ietf.org/doc/html/rfc9110#section-12.4.2
+//   - https://developer.mozilla.org/en-US/docs/Glossary/Quality_values
+//
+// For the following header value:
 //
 //	"text/html,text/*;q=0.5,*/*;q=0.7"
 //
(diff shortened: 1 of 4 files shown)


util/httputil/httputil.go
```diff
@@ -77,3 +65,23 @@ func ParseMultiValuesHeader(header string) []HeaderValue {
...
+// by RFC 9110, and 0 if the quality value cannot be parsed.
+func parseQuality(params string) float64 {
+	for param := range strings.SplitSeq(params, ";") {
+		name, value, ok := strings.Cut(param, "=")
+		if !ok || !strings.EqualFold(strings.TrimSpace(name), "q") {
+			continue
+		}
+		q := strings.TrimSpace(value)
+		priority, err := strconv.ParseFloat(q, 64)
+		if err != nil || !qualityValueRegex.MatchString(q) {
```
System-Glitch: I think we can make it even faster by completely removing the regex and just counting how many characters there are after the dot (if there is a dot). At this point we know it's a proper number since ParseFloat didn't return an error.

util/httputil/httputil_test.go
```diff
@@ -53,3 +53,73 @@ func TestParseMultiValuesHeader(t *testing.T) {
 	result = ParseMultiValuesHeader("   ")
 	assert.Equal(t, expected, result)
 }
+
+func TestParseMultiValuesHeaderQualityValue(t *testing.T) {
```
System-Glitch: Thanks for the added tests. I think we should merge the two tests together. The old one isn't even using a test case structure...

util/httputil/httputil_test.go
```diff
@@ -53,3 +53,73 @@ func TestParseMultiValuesHeader(t *testing.T) {
...
+			header: "text/html;",
+			want:   []HeaderValue{{Value: "text/html", Priority: 1}},
+		},
+		{
+			desc:   "not_a_number",
+			header: "gzip;q=abc, br;q=0.1",
+			want:   []HeaderValue{{Value: "br", Priority: 0.1}, {Value: "gzip", Priority: 0}},
+		},
+		{
+			desc:   "out_of_range",
```
System-Glitch, line 129: Please add a case for q < 0.

util/httputil/httputil_test.go
```diff
@@ -53,3 +53,73 @@ func TestParseMultiValuesHeader(t *testing.T) {
...
 }
+
+func TestParseMultiValuesHeaderQualityValue(t *testing.T) {
+	cases := []struct {
+		desc   string
+		header string
+		want   []HeaderValue
+	}{
+		{
+			desc:   "whitespace_before_quality_value",
```
System-Glitch, line 75: Please add a case where there are spaces around the value: `gzip;q= 0.9, br;q=0.8 , bz;q = 0.7 `

util/httputil/httputil.go
```diff
@@ -14,14 +14,18 @@ type HeaderValue struct {
...
-var qualityValueRegex = regexp.MustCompile(`^q=([01]\.[0-9]{1,3})$`)
+var qualityValueRegex = regexp.MustCompile(`^[01](\.[0-9]{1,3})?$`)
 
 // ParseMultiValuesHeader parses multi-values HTTP headers, taking the
 // quality values into account. The result is a slice of values sorted
 // according to the order of priority.
 //
 // The input is trimmed. If the input is empty, returns an empty slice.
 //
+// The priority of a value is the value of its "q" parameter. Values without
```
System-Glitch: Please add the RFC 9110 reference in the documentation too, now that it properly complies with it.

I think we should clarify "The input is trimmed." too by adding details about trimming the value and q as well.

Lastly, it should also state that other parameters are supported (e.g.: charset).

I can do it after this PR is merged.

System-Glitch (changes requested):

System-Glitch:
About the drawbacks: this behavior change is acceptable to me on account that the previous implementation was not compliant with the standard and didn't work as one would expect. The previous implementation was thus buggy. 

It is not a breaking change because applications will keep working and compiling without any developer action needed. The concerns are valid though, some clients may be hit but I don't expect this to be a real problem at all; the impact is very minor.

System-Glitch:
Thanks for the changes, I'll have a look later this week.

System-Glitch:
I added a couple test cases and reworded the docs in some places. LGTM and thanks a lot for this contribution!

System-Glitch (approved):

System-Glitch:
Squashed and merged, thanks! 🎉 This will be published as v5.12.4.

PR #296, opened 2026-09-15
compress: fix quality value of 0 being ignored in getEncoder

MaxFreedomPollard:
## References

**Issue(s):** N/A  
**Discussion:** N/A

## Description

`Middleware.getEncoder` (`middleware/compress/compress.go:144` on master) groups the parsed `Accept-Encoding` values by priority at line 152 and matches the configured encoders against each group, but it never compares a priority to 0. RFC 9110 section 12.4.2 defines a quality value of 0 as "not acceptable", and section 12.5.3 applies that rule to `Accept-Encoding`: a listed content coding is acceptable unless it is accompanied by a qvalue of 0. The middleware therefore compresses responses with an encoding the client explicitly refused.

With `Encoders: []Encoder{&Gzip{Level: gzip.BestCompression}}`:

- `Accept-Encoding: gzip;q=0` comes back with `Content-Encoding: gzip` and a gzip framed body. A client that sends `q=0` for gzip is exactly the client that cannot decode gzip, so the body is unreadable to it.
- `Accept-Encoding: *;q=0` does the same, even though it says no content coding is acceptable at all.
- `Accept-Encoding: gzip;q=0, *` also returns gzip, because the wildcard branch at line 165 returns `m.Encoders[0]` unconditionally. The wildcard matches the codings that are not explicitly listed, so gzip is the one encoding it must not pick here.

That same unconditional `m.Encoders[0]` panics with `index out of range [0] with length 0` when `Encoders` is empty and the request sends `Accept-Encoding: *`. Every other header value already returns nil for an empty `Encoders` slice, so only the wildcard group reaches the panic.

`httputil.ParseMultiValuesHeader` already parses the quality value correctly since #295 and returns a `Priority` of 0 for the `q=0` entries in the first three headers. Only the consumer was missing.

The fix splits the parsed values before any matching. They come back sorted by descending priority, so the zero-priority values are always a suffix: `slices.IndexFunc` locates the first one, everything before it is accepted and everything from it onwards is rejected. When nothing is accepted, `getEncoder` returns nil and the middleware passes the response through uncompressed. The wildcard branch now returns the first configured encoder that is not in the rejected list, and nil when every configured encoder was rejected or when `Encoders` is empty, which removes the panic. The priority ordering, the rule that the earliest encoder in the slice wins on equal priority, and the hijack and `Upgrade` short circuits are untouched. The doc comment on `Middleware` now describes the rejection rule and links both RFC sections.

Thirteen cases were added to the existing `TestEncoderPriority` table: `gzip;q=0`, `gzip;q=0.000`, `gzip; Q = 0`, `gzip;q=0.001`, `gzip;q=0, br`, `br;q=0, zstd;q=0, gzip;q=0`, `*;q=0`, `*;q=0, gzip`, `br;q=0, *`, `br;q=0, *;q=0.5`, `br;q=0, zstd;q=0, gzip;q=0, *`, `gzip;q=0, *` with a single gzip encoder, and `*` with an empty `Encoders` slice. A `Rejected encoding` subtest in `TestCompressMiddleware` checks that a full request with `Accept-Encoding: gzip;q=0` returns the plain body and no `Content-Encoding` header. Ten of the table rows and that subtest fail on master, the empty `Encoders` row by panicking at `compress.go:165`; the other three rows pass on master and guard behavior that does not change.

`go test -race -coverprofile=coverage.txt -covermode=atomic -coverpkg=./... ./middleware/compress/...` passes and `go tool cover -func` reports `getEncoder` at 100% of statements. `gofmt -s -l .` prints nothing and `go vet ./middleware/compress/...` is clean. golangci-lint was not run locally.

### Possible drawbacks

The headline change is a behavior change: a response to `Accept-Encoding: gzip;q=0` used to be gzip encoded and is now sent uncompressed. That is the intended change, RFC 9110 section 12.5.3 requires it, no exported signature changes, and applications keep compiling and working without developer action.

Two further header shapes change with it.

`parseQuality` (`util/httputil/httputil.go:85`) returns 0 both for an explicit `q=0` and for a quality value it cannot parse, and its doc comment says so. A request sending `Accept-Encoding: gzip;q=abc`, `gzip;q=1e-2` or an out of range value such as `gzip;q=-1` used to be compressed at lowest priority and is now not compressed.

The wildcard part is a separate change from the plain `q=0` part. `Accept-Encoding: gzip;q=0, *` used to select the first encoder in the slice and now skips the rejected ones. RFC 9110 section 12.5.3 justifies it independently: the wildcard matches the codings that are not listed in the field, and gzip is listed there. The change only excludes codings the header rejected with `q=0`; a coding listed with a non-zero quality value can still be selected through the wildcard, as before. `lang.Languages.DetectLanguage` (`lang/lang.go:166`) reads the same parser and has the same `q=0` gap for `Accept-Language`; I left it out of this change to keep it to the compress middleware.


middleware/compress/compress.go
```diff
@@ -145,7 +159,16 @@ func (m *Middleware) getEncoder(response *goyave.Response, request *goyave.Reque
...
 		return nil
 	}
-	acceptedEncodings := httputil.ParseMultiValuesHeader(request.Header().Get("Accept-Encoding"))
+	values := httputil.ParseMultiValuesHeader(request.Header().Get("Accept-Encoding"))
+
+	// A quality value of 0 means the encoding is not acceptable
+	// (RFC 9110 sections 12.4.2 and 12.5.3).
+	// The values are sorted by descending priority, so the rejected ones are all at the end.
+	acceptedEncodings := values
+	var rejectedEncodings []httputil.HeaderValue
```
System-Glitch: You could simplify by just filtering out the `values` that have a priority of 0.

```go
acceptedEncodings := lo.Filter(values, func(v httputil.HeaderValue, _ int) bool { return v.Priority != 0 })
```

With this the rest of the changes are not  necessary and the implementation is easier to understand in my opinion.

middleware/compress/compress_test.go
```diff
@@ -336,6 +353,71 @@ func TestEncoderPriority(t *testing.T) {
...
+			acceptEncoding: "br;q=0, zstd;q=0, gzip;q=0, *",
+			want:           nil,
+		},
+		{
+			encoders:       []Encoder{gzip},
+			acceptEncoding: "gzip;q=0, *",
+			want:           nil,
+		},
+		{
+			encoders:       []Encoder{},
```
System-Glitch, line 417: To fix this panic, I would simply add an early-return if `m.Encoders` is empty.

System-Glitch (changes requested):
Thanks again for this PR. Compliance with standards is very important to me.

As explained in the PR description, the [language detection](https://github.com/go-goyave/goyave/blob/42777290c51f93487f2278d8bd3bf06090f2c468/lang/lang.go#L165) is incorrect as well. A simple fix would be to add the following and a quick test case:

```go
if lang.Priority == 0 { // Not acceptable according to RFC 9110
	continue
}
```
I'd like to bundle this change in the MR as well so we fix all compliance problems in one patch release.

I would add a small comment to `httputil.ParseMultiValuesHeader` to tell users that a priority of 0 means that it's not acceptable too.

MaxFreedomPollard:
Pushed `e5284e3`. `getEncoder` filters with `lo.FilterReject` and returns early on empty `Encoders`. I kept the rejected list for the wildcard only: with the filter alone, `br;q=0, *` falls through to `Encoders[0]`, which is br, and three table rows cover that. Happy to drop it with those rows if you prefer. `DetectLanguage` skips `q=0` (four new cases), and the `ParseMultiValuesHeader` doc now says what priority 0 means.

lang/lang_test.go
```diff
@@ -174,6 +174,12 @@ func (suite *LangTestSuite) TestDetectLanguage() {
 	suite.Equal(l.languages["en-US"], l.DetectLanguage("*"))
 	suite.Equal(l.languages["en-US"], l.DetectLanguage("notalang"))
 	suite.Equal(l.languages["en-US"], l.DetectLanguage(""))
+
+	// A quality value of 0 means "not acceptable"
+	suite.Equal(l.languages["en-US"], l.DetectLanguage("fr-FR;q=0"))
+	suite.Equal(l.languages["en-US"], l.DetectLanguage("fr;q=0"))
+	suite.Equal(l.languages["en-US"], l.DetectLanguage("notalang, fr-FR;q=0"))
+	suite.Equal(l.languages["fr-FR"], l.DetectLanguage("en-US;q=0, fr-FR;q=0.1"))
```
System-Glitch, line 182: Please add a test case for the default language fallback if no other language matches:
```go
suite.Equal(l.languages["en-US"], l.DetectLanguage("en-US;q=0"))
```

System-Glitch (approved):
Thank you for the changes and additional docs and tests, that's awesome. 👍🏻 

> I kept the rejected list for the wildcard only: with the filter alone, br;q=0, * falls through to Encoders[0], which is br, and three table rows cover that. Happy to drop it with those rows if you prefer. 

I didn't think about this specific case at first; hence why I thought at first that the rejected list wasn't necessary. But that's good we should keep that.

Currently a bit busy but I'll merge that soon. Thanks a lot for your contribution!

lang/lang_test.go
MaxFreedomPollard: Added in `84821b4`, thanks.

